PR #5, opened 2021-05-28
readme: make the helm command compatible with zsh

maelvls:
The { and } have a different meaning between bash and zsh, let's
quote the entire blob.


JoshVanL:
Closing as I don't think this is relevant any more.

/close

PR #11, opened 2021-08-11
Refactor to internal and use a evaluator registry

JoshVanL:
Branched from #7 which should be merged first.

This PR introduces a new `/registry` package which is intended to be used by both the base and external policy approvers to register their policy evaluations.

This PR also moves most things into `/internal`. This makes things easier for external policy consumers, and prevents them from being tempted to re-use or import things they shouldn't be.

External policy evaluators need only import `/cmd/app` and implement a simple main package.

The main policy evaluator has been renamed to manager, which now operates over the registry evaluators.

The base policy evaluator has been renamed to `base` and is registered via the `/registry`.

Hopefully most of this PR is file renaming, but can be reviewed commit by commit if it's unmanageable.

Renames `base` approver to `attribute`.

/assign @charlieegan3


internal/pkg/base/base.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package policy
+package base
```
charlieegan3, line 17: suggestion: I still see `base` as a misleading name - can we think of a clearer one? If we're going to have many pluggable options, and this is just one, I'd rather not have a name that suggests (to me at least) that this is foundational (like a buttery biscuit base is foundational to a cheesecake... 😋) (i.e. that the others might be somehow dependent on it or be layered on top)

Naming is hard, but some ideas might be:
- 'rule based'
- 'attribute based/validation'
- 'property based/validation'
- ...

Just a thought.

JoshVanL: Yeah I understand where you are coming from. I would argue though that this is the base evaluator. All evaluators operate over CertificateRequestPolicies, and base is responsible for the default evaluations.

charlieegan3: In that case, would `default` be more accurate than `base`?

JoshVanL: Yeah true, just feel `default` suggests that it is optional or should be replaced with a non-default approver which it is not.

Not wedded to the name though :shrug:

charlieegan3: I think the main concern I have is that base suggests that other approvers might be layered on top - which is also not the case.

You're not keen on something like `attribute` or `property`?

JoshVanL: `attribute` works for me :+1:

charlieegan3: 👍 thanks, I hope this will be clearer.

PR #13, opened 2021-08-13
Refactor apis

JoshVanL:
Branched from #11 

More house cleaning.

Moves policy api group into directory.

Make code-gen binaries as a depend, refactor Makefile.

Not much interesting here :)

PR #14, opened 2021-08-14
Refactor tests

JoshVanL:
Branched from #13 which should be merged first

This PR does away with the e2e tests, and moves them into a locally run tests using controller-runtime envtest.

This makes running tests for policy-approver, and future plugins, much quicker as we don't have to wait for a kind cluster etc.

This does away with the e2e tests. Later on, we will replace it with a smoke test which will install policy-approver against a kind cluster. Proper testing should always be done against `/test/env`.

`/test/env` is designed to be reusable by other projects.

`/test/env` will setup a API server, installs the CRDs, and starts the cert-manager webhook.

/assign @charlieegan3


internal/pkg/controller/test/tests.go
```diff
@@ -0,0 +1,296 @@
...
+
+import (
+	"context"
+	"crypto/x509"
+
+	apiutil "github.com/jetstack/cert-manager/pkg/api/util"
+	cmapi "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
+	cmmeta "github.com/jetstack/cert-manager/pkg/apis/meta/v1"
+	"github.com/jetstack/cert-manager/test/unit/gen"
+	. "github.com/onsi/ginkgo"
```
charlieegan3, line 27: question: are we using ginkgo for more than before/after functions ('suites') and `Expect` functionality?

question: do we use ginkgo in other Jetstack projects?

charlieegan3: I see it's in the CM e2e tests. In JSCP we use testify suite which I'd say is more idiomatic for test suite functionality. I suppose the waiting functionality is another benefit?

Not challenging it being here, but interested to know the rationale since I think something similar in testify could be easier to use for anyone coming from JSCP. ginkgo has some unconventional syntax it seems.

charlieegan3:
/lgtm
/hold

Hold added to allow response to above comment, please feel free to merge after that.

internal/pkg/controller/test/tests.go
JoshVanL: @charlieegan3 

> question: are we using ginkgo for more than before/after functions ('suites') and Expect functionality?

Yeah it has some more useful things like `Eventually` etc. We are also exporting the test results to `/_artifacts` which get pushed to a google bucket in our e2e.

Optional concurrency is also nice.

> question: do we use ginkgo in other Jetstack projects?

We use it for most if not all cert-manager projects.

---

Not particularly wedded to using ginkgo, we can remove it if you feel it will become a nuisance.

charlieegan3: 👍 thanks.

PR #19, opened 2021-08-16
`IssuerRefSelector`

JoshVanL:
Branched from #18 which should be merged first.

This PR introduces the Webhook as a member of Approver. Webhook is responsible for implementing admission functions for that approver. Shared webhook server exposes these funcs to kube.

`AllowedIssuers` has been changed to `IssuerRefSelector` which is used to match over CertificateRequests. If no policies are bound or matched against a request then it is unprocessed. Tests have been updated to reflect this.

`IssuerRefSelector` is a wildcard string matcher. The following `IssuerRefSelector` would match all `IssuerRefs`:

```
issuerRefSelector: {}
```

I haven't updated the `README.md` since I want to do all the docs in a separate PR.

I want to update the controller integration tests in a later PR to extend its coverage.

/assign @charlieegan3


irbekrm:
/assign

I'll review this once #18 gets merged and this rebased

JoshVanL:
@irbekrm This should be good for another look now

PR #20, opened 2021-08-17
Approver plugins

JoshVanL:
Branched from #19 which should be merged first.

This PR introduces `spec.plugins` on the CertificateRequestPolicy. This field is used to configure the active plugins for a policy, and the related configuration it needs.

Registered plugins and then make use of their entry on a policy to evaluate CertificateRequests.

I've added some more funcs to the Approver interface so that plugins can be fully featured. No two plugins may be registered with the same name.

/assign @charlieegan3


JoshVanL:
:raised_eyebrow: 
/retest

PR #25, opened 2021-08-20
Smoke test

JoshVanL:
Branched from #24 which should be merged first

Adds a smoke test which will spin up a kind cluster, install cert-manager with the default approver off, install policy-approver, run some sanity tests.

Before this is merged, we need to add a smoke-test target to prow but not until the previous PRs have been merged.

/assign @irbekrm 
/hold


JoshVanL:
The smoke test running needs to use the golang-dind image, but 1.17 is not available yet.

irbekrm:
Should be now eu.gcr.io/jetstack-build-infra-images/golang-dind:latest-1.17

irbekrm:
Its using the wrong image

irbekrm:
It'll need docker set up https://github.com/jetstack/testing/pull/577

Hopefully these will go away with new PRs https://github.com/cert-manager/approver-policy/pull/25#issuecomment-914589120

irbekrm:
That was still the old config https://prow.build-infra.jetstack.net/prowjob?prowjob=c8b17c7e-1016-11ec-afd1-c264214494a5

irbekrm:
It's the same networking error it seems, from pod logs:
```
Set kubectl context to "kind-policy-approver"
You can now use your cluster with:

kubectl cluster-info --context kind-policy-approver

Not sure what to do next? 😅  Check out https://kind.sigs.k8s.io/docs/user/quick-start/
>> loading docker image...
Image: "quay.io/jetstack/cert-manager-policy-approver:smoke" with ID "sha256:4b25b33fd4b620969a32da10c69feda98a2d4f58c349f6a88cea15bdfc0d99b2" not yet present on node "policy-approver-control-plane", loading...
>> installing cert-manager...
Error: looks like "https://charts.jetstack.io" is not a valid chart repository or cannot be reached: Get "https://charts.jetstack.io/index.yaml": dial tcp: lookup charts.jetstack.io on 10.8.240.10:53: read udp 10.8.2.15:47140->10.8.240.10:53: i/o timeout
make: *** [Makefile:67: demo] Error 1
Cleaning up after docker in docker.
```

/retest

test/smoke/tests.go
```diff
@@ -0,0 +1,171 @@
...
+
+		By("Creating test SelfSigned Issuer")
+		issuer := cmapi.Issuer{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "smoke-test-policy-",
+				Namespace:    namespace.Name,
+			},
+			Spec: cmapi.IssuerSpec{IssuerConfig: cmapi.IssuerConfig{SelfSigned: &cmapi.SelfSignedIssuer{}}},
+		}
+		Expect(cl.Create(ctx, &issuer)).NotTo(HaveOccurred())
```
irbekrm, line 64: So because we don't test that a cert gets issued we therefore don't need to wait for the Issuer to become Ready 👍🏼

hack/ci/delete-cluster.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
irbekrm, line 1: This isn't used in CI it seems, but could be useful when running locally I guess 👍🏼

irbekrm (approved):
This makes sense and seems like a very good and non-controversial addition 👍🏼 

/lgtm

PR #43, opened 2021-11-09
Support helm releases in non-"cert-manager" namespaces.

### LINKED ISSUE #42: Hardcoded namespace in code
Thank you for this great project.

Adding it to cert-manager caused an issue running the solution in a namespace other then cert-manager.
The TLS connection fails due to an incorrect SAN name in the generated certificate for the webhook.
Investigation showed an hardcoded DNS Name in go package pkg/internal/webhook/tls/tls.go

        template := &x509.Certificate{
                Version:            2,
                PublicKeyAlgorithm: x509.ECDSA,
                PublicKey:          pk.Public(),
                DNSNames:           []string{"**cert-manager-approver-policy.cert-manager.svc**"},
                KeyUsage:           x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment,
                ExtKeyUsage:        []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
        }

"cert-manager-approver-policy.cert-manager.svc" should be parameterised to be able to use in an alternative namespace.

JoshVanL:
Adds support for releasing approver-policy is a different namespace by
exposing the service name of the webhook, and piping the secret
namespace to the tls manager.

This PR is a no-op except approver-policy will now work when releasing in a namespace that is not cert-manager.

/assign @irbekrm 

fixes #42

### DIFF
--- a/deploy/charts/approver-policy/templates/deployment.yaml
+++ b/deploy/charts/approver-policy/templates/deployment.yaml
@@ -30,17 +30,19 @@ spec:
           periodSeconds: 7
         command: ["cert-manager-approver-policy"]
         args:
-          - "--log-level={{.Values.app.logLevel}}"
+          - --log-level={{.Values.app.logLevel}}
+
           {{- range .Values.app.extraArgs }}
-          - "{{ . }}"
+          - {{ . }}
           {{- end  }}
 
-          - "--metrics-bind-address=:{{.Values.app.metrics.port}}"
-          - "--readiness-probe-bind-address=:{{.Values.app.readinessProbe.port}}"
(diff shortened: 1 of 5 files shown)


irbekrm (approved):
I've tested deploying in another namespace and creating some resources- all worked as expected 👍🏼 

Now when the namespaces are properly configurable wondering if it would ever be acceptable or needed for any scenario to deploy multiple approver-policy's in a cluster. (I am guessing that no, since there would be no way how to have one of them to only look at `CertificateRequest`s in a particular namespace.)

/lgtm

pbloed:
Hi Josh,

Are you going to create a new release any time soon?
Would be nice if fixes are also released.

PR #44, opened 2021-11-10
Adds log as a parameter to the Approver Prepare signature.

JoshVanL:
This helps approvers to use a consistent logger which is a child of the main
approver-policy manager.

/assign @wallrj

### DIFF
--- a/pkg/approver/allowed/allowed.go
+++ b/pkg/approver/allowed/allowed.go
@@ -19,6 +19,7 @@ package allowed
 import (
 	"context"
 
+	"github.com/go-logr/logr"
 	"github.com/spf13/pflag"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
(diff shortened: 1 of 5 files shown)


wallrj (approved):
Thanks @JoshVanL 

I personally would prefer the logger to be added to the context of Prepare, Ready, etc.
so that it it would be easier to view all the log messages concerning a single Reconcile loop.
approver-policy could add a unique UUID to the logger when it first begins reconciling, and pass that (via the context) to all the plugins.
This way in the case of a Reconcile failure you could trace all the messages leading up to the failure.

But I know we discussed this at a previous standup, and you have reasonable objections to adding stuff to the context, which I kind of agree with too.

Unhold if you prefer to merge this and make progress.

/lgtm
/hold

JoshVanL:
Thanks @wallrj- I do still hold the view of not wanting to put the logger in the context.

I'd have to give some more thought as to whether it would be a good idea to add a logger as a _separate_ parameter to `Validate()`, `Reconcile()` etc. While true, a logger with a UUID for each approver invocation from the manager does provide a better story in the logs, I think I would like to see a really clear example where `1.` the logged information isn't suitable to be returned as a return argument back to the manager, and `2.` the name of the policy [and request] is not enough identifiable information to complete the story in the logs.

/hold cancel

PR #46, opened 2021-11-22
Controller Certificaterequests RBAC

### LINKED ISSUE #45: Approver-policy does not reconcile CRs on RBAC updates
I was manually deploying cert-manager and approver-policy resources and made a mistake in RBAC, so the `CertificateRequest` ended up being 'unprocessed' (no matching policy). I corrected the RBAC resources, but the CR was not reconciled and I had to delete it after which the new one got correctly approved.

Would it make sense for the [certificaterequests controller](https://github.com/cert-manager/approver-policy/blob/main/pkg/internal/controllers/certificaterequests.go#L71) to reconcile unprocessed `CertificateRequest`s on related (Cluster)Role/(Cluster)RoleBinding events similarly to how it's currently done on `CertificateRequestPolicy` events?

Steps to reproduce:
1. Create an `Issuer`, `Certificate` and `CertificateRequestPolicy` matching the issuer (but no RBAC)
2. Observe in approver-policy logs that no matching policy was found and the CR is neither approved nor denied
3. Create the RBAC
4. Observe that the `CertificateRequest` still does not get approved/denied
5. Delete the `CertificateRequest` and observe the new one get correctly approved

JoshVanL:
Fixes #45 

This PR changes the CertificateRequest controller to now reconcile on `Role`, `RoleBinding`, `ClusterRole`, and `ClusterRoleBinding` events. This means that all non Approved or Denied request will be re-reconciled if any RBAC resource is created/updated/deleted. If any resource changes whether a request is appropriate for a policy, that request will then be evaluated.

/assign @irbekrm

### DIFF
--- a/deploy/charts/approver-policy/templates/clusterrole.yaml
+++ b/deploy/charts/approver-policy/templates/clusterrole.yaml
@@ -16,6 +16,7 @@ rules:
 - apiGroups: ["cert-manager.io"]
   resources: ["certificaterequests"]
   verbs: ["list", "watch"]
+
 - apiGroups: ["cert-manager.io"]
   resources: ["certificaterequests/status"]
   verbs: ["update"]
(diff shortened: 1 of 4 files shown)


irbekrm:
I tested this and verified that it fixes #45 . The code looks good too 👍🏼 

Thanks for fixing it!

/lgtm
/approve

PR #47, opened 2021-11-23
Upgrade cert-manager to v1.6.1

JoshVanL:
This PR updates the cert-manager go module to use `v1.6.1`.

As part of updating to that version, the [`StartWebhookServer`](https://github.com/jetstack/cert-manager/blob/5eabeb0020906f741da05122e4dee62d2d952657/cmd/webhook/app/testing/testwebhook.go#L60) func now takes a context as an extra argument. To account for this `RunControlPlane` now also takes a context, which is created in all the consuming locations. 

Consumers of this repo and the `test/env` package will also need to update their signatures to match.

/assign @SgtCoDFish


pkg/approver/manager/predicate/predicate_test.go
```diff
@@ -36,8 +36,13 @@ import (
 )
 
 func Test_RBACBound(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.TODO())
```
irbekrm, line 39: So we're using the functionality that Tim added in https://github.com/jetstack/cert-manager/pull/4335 to ensure that the webhook server gets stopped when these tests finish 👍🏼

irbekrm (approved):
/lgtm

PR #59, opened 2022-05-03
Adds verify-helm-docs.sh and update-helm-docs.sh scripts. Verify helm-docs during lint

JoshVanL:
/assign @wallrj 

This PR adds a scripts, run at make lint, which verifies that the helm chart README.md file is kept up to date.

An update script is also added which will update the helm chart README.md.


wallrj (approved):
Thanks @JoshVanL 

I tested it and it worked.

FYI There's an open feature request about making helm-docs work with commented out sub-fields:
 * https://github.com/norwoodj/helm-docs/issues/98

/lgtm

PR #72, opened 2022-06-09
Creates a go.mod in hack/tools

JoshVanL:
Creates a `go.mod` in `hack/tools` so that binary dependencies have a seperate `go.mod` so to not conflict with the main project.

Also updates both `go.mod` which because of this change, in theory fixes the problems we have been seeing with cert-manager and other projects being downgraded.

/assign @wallrj


wallrj:
The builds failed with errors like:

```
 go build -o /home/prow/go/src/github.com/cert-manager/approver-policy/bin/deepcopy-gen k8s.io/code-generator/cmd/deepcopy-gen
missing go.sum entry for module providing package k8s.io/code-generator/cmd/deepcopy-gen; to add:
	go mod download k8s.io/code-generator 
```

I took a different approach and used `go install` in #75 . See what you think.

JoshVanL:
@wallrj Please take another look

PR #74, opened 2022-06-10
House Cleaning: Move manager implementation into /pkg/internal

JoshVanL:
The main crux on this PR is some house cleaning to move more things into `internal`.

The approver manager implementation has been moved into `internal`.

The `pkg/approver/internal` wildcard implementation has been moved into `pkg/internal/util`.

The core approvers (`allowed` and `constraints`) have their structs made private, and moved into `pkg/internal/approver`.

/assign @wallrj


wallrj (approved):
Thanks @JoshVanL 

Seems like an improvement to  make more of the code private for now.
And I checked my approver-policy TPP plugin and it doesn't use these now private APIs.

/lgtm

Issue #62, opened 2022-05-20
CertificateRequestPolicy based on which namespace the certificate request belongs to


smuda:
My use case is to only allow certificate requests with dnsName which are service-names from the namespace to which the service-name belongs to, for a certain cluster-issuer. For example the dnsName `my-service.my-space.svc` should only be allowed when requested from the "my-space" namespace.

Preferably, I'd also like to restrict namespaces for a certain cluster issuer.

I don't seem to find how to do that in the current implementation. Am I missing something or is this an uncommon usecase?

rickymulder:
Is this issue still valid?  Looks like it was resolved in #96 
To accomplish this, there is a fairly complex rbac setup, but in order for a serviceaccount to make use of a request policy it must have the "use" action for the exact policy name.

smuda:
Ops, I hadn't seen the PR. Let me check it out.

sastorsl:
I would like to see an option where I can specify a fixed policy for a number of namespaces, without having to create a policy for all those namespaces.

An example would be that the `allowed` list could have placeholders.

Example policy:

```yaml
---
apiVersion: policy.cert-manager.io/v1alpha1
kind: CertificateRequestPolicy
metadata:
  name: example-com
spec:
  allowed:
    # Name TBD - trigger that a <namespace> placeholder have to match the requesting namespace
    namespace_restricted: True
    commonName:
      value: "*.<namespace>.example.com"
    dnsNames:
      values:
      - "*.<namespace>.example.com"
    usages:
    - "server auth"
  selector:
    issuerRef:
      name: le-issuer
      kind: ClusterIssuer
      group: cert-manager.io
```

Should be allowed:

```yaml
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: my-certificate
  namespace: my-test
spec:
  dnsNames:
  - my-certificate.my-test.example.com
  issuerRef:
    kind: ClusterIssuer
    name: le-issuer
  secretName: my-certificate
```

Should be rejected

```yaml
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: my-certificate
  namespace: my-othernamespace  # Namespace changed
spec:
  dnsNames:
  - my-certificate.my-test.example.com
  issuerRef:
    kind: ClusterIssuer
    name: le-issuer
  secretName: my-certificate
```

This would prevent:

* copy/paste errors
* One namespace grabbing or overwriting a hostname for one namespace into the other

Would achieve:

[...]

PR #96, opened 2022-07-11
CertificateRequestPolicy: selector.namespace

JoshVanL:
Related #62 

See [here](https://github.com/cert-manager/website/pull/1032) for documentation PR.

This PR adds `selector.namespace` to the CertificateRequestPolicy, which enables another option for selecting over CertificateRequests which should be evaluated by a policy.

`selector.namespace` allows for setting `matchNames` and `matchLabels`.

At least one of `issuerRef` or `namespace` selectors must be defined. Both can be defined at the same time.

```yaml
apiVersion: policy.cert-manager.io/v1alpha1
metadata:
  name: my-policy
spec:
  ...
  selector:
    namespace:
      matchNames:
      - "default"
      - "app-team-*"
    matchLabels:
      foo: bar
      team: dev
```

```yaml
apiVersion: policy.cert-manager.io/v1alpha1
metadata:
  name: match-all-requests
spec:
  ...
  selector:
    namespace: {}
```


deploy/charts/approver-policy/templates/crds/policy.cert-manager.io_certificaterequestpolicies.yaml
```diff
@@ -280,13 +280,11 @@ spec:
...
                           type: string
                 selector:
-                  description: Selector is used for selecting over which CertificateRequests this CertificateRequestPolicy is appropriate for and so will used for its evaluation.
+                  description: Selector is used for selecting over which CertificateRequests this CertificateRequestPolicy is appropriate for and so will used for its approval evaluation.
                   type: object
-                  required:
-                    - issuerRef
                   properties:
                     issuerRef:
-                      description: "IssuerRef is used to match this CertificateRequestPolicy against processed CertificateRequests. This policy will only be evaluated against a CertificateRequest whose `spec.issuerRef` field matches `spec.selector.issuerRef`. CertificateRequests will not be processed on unmatched `issuerRef`, regardless of whether the requestor is bound by RBAC. Accepts wildcards \"*\". Nil values are equivalent to \"*\", \n The following value will match _all_ `issuerRefs`: ``` issuerRef: {} ``` \n Required field."
```
munnerz, line 289: Are `*` characters supported anywhere in the namespace name? What about multiple wildcards? (e.g. `*-prod-*`)?

Also here you mention `nil values` - but from an end-user of the API perspective it isn't clear what "Nil" means IMO - can we change this to be `the empty value ("")` or something like that?

deploy/charts/approver-policy/templates/crds/policy.cert-manager.io_certificaterequestpolicies.yaml
```diff
@@ -298,6 +296,41 @@ spec:
...
+                                type: array
+                                items:
+                                  type: string
+                        matchLabels:
+                          description: matchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels map is equivalent to an element of matchExpressions, whose key field is "key", the operator is "In", and the values array contains only "value". The requirements are ANDed.
+                          type: object
+                          additionalProperties:
+                            type: string
+                        names:
+                          description: MatchNames are the set of Namespace names that select on CertificateRequests that have been created in a matching Namespace. Accepts wildcards "*".
```
munnerz, line 330: Here you mention `MatchNames` but the field is called `names`

deploy/charts/approver-policy/templates/crds/policy.cert-manager.io_certificaterequestpolicies.yaml
munnerz: Further to that, if they _are_ meant to be supported anywhere in the namespace name, can we get corresponding test cases?

JoshVanL: > Are * characters supported anywhere in the namespace name? What about multiple wildcards? (e.g. *-prod-*)?

Yes, [see](https://cert-manager.io/docs/projects/approver-policy/#allowed) 

```
Allowed string fields accept wildcards "*" within its values. Wildcards "*" in
patterns represent any string that has a length of 0 or more. A pattern
containing only "*" will match anything. A pattern containing `"*foo"` will
match `"foo"` as well as any string which ends in `"foo"` (e.g. `"bar-foo"`). A
pattern containing `"*.foo"` will match `"bar-123.foo"`, but not `"barfoo"`.
```

> Also here you mention nil values - but from an end-user of the API perspective it isn't clear what "Nil" means IMO - can we change this to be the empty value ("") or something like that?

Yep agree :+1:

JoshVanL: > Further to that, if they are meant to be supported anywhere in the namespace name, can we get corresponding test cases?

Sure, more than happy to add more test cases for this and will, but will mention that the actual wildcard implementation is already well tested. https://github.com/cert-manager/approver-policy/blob/main/pkg/internal/util/wildcard_test.go

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -331,6 +341,23 @@ type CertificateRequestPolicySelectorIssuerRef struct {
...
+	// MatchNames are the set of Namespace names that select on
+	// CertificateRequests that have been created in a matching Namespace.
+	// Accepts wildcards "*".
+	// +optional
+	MatchNames []string `json:"names,omitempty"`
+
+	// LabelSelector is the set of Namespace labels that select on
+	// CertificateRequests that have been created in a matching Namespace.
+	// +optional
+	*metav1.LabelSelector `json:",omitempty"`
```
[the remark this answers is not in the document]
JoshVanL: Yes, I think we shouldn't be using the whole `metav1.LabelSelector` here. Will opt for changing this to a `MarchLabels map[string]string \`json:"matchLabels",omitempty"\``, and building the label selector that way.

JoshVanL:
Thanks for the review @munnerz, PTAL

pkg/internal/approver/manager/predicate/predicate.go
```diff
@@ -77,6 +86,64 @@ func SelectorIssuerRef(_ context.Context, cr *cmapi.CertificateRequest, policies
...
+				matchingPolicies = append(matchingPolicies, policy)
+				continue
+			}
+
+			ns := cr.Namespace
+			matched := len(nsSel.MatchNames) == 0
+
+			// Match by name.
+			for _, matchName := range nsSel.MatchNames {
+				if util.WildcardMatchs(matchName, ns) {
```
munnerz, line 118: `WildcardMatches` (missing an `e`)

pkg/internal/approver/manager/predicate/predicate.go
```diff
@@ -77,6 +86,64 @@ func SelectorIssuerRef(_ context.Context, cr *cmapi.CertificateRequest, policies
...
+					matched = true
+					break
+				}
+			}
+			if !matched {
```
munnerz, line 123: It makes sense why you have the `matched` assignment checking the `len` for `0` here now, but it has taken me a bit of head wrangling.. maybe a comment above the initial `matched` declaration would help to explain what's going on here?

pkg/internal/approver/manager/predicate/predicate.go
```diff
@@ -77,6 +86,64 @@ func SelectorIssuerRef(_ context.Context, cr *cmapi.CertificateRequest, policies
...
+				continue
+			}
+
+			// Match by Label Selector.
+			if nsSel.MatchLabels != nil {
+				selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
+					MatchLabels: nsSel.MatchLabels,
+				})
+				if err != nil {
+					return nil, fmt.Errorf("failed to parse namespace label selector: %w", err)
```
munnerz, line 133: How does this error get handled further up the chain? Can we perform validation on the map values to avoid errors here? I'm sure there's a regex somewhere in the k8s codebase for this

pkg/internal/approver/manager/predicate/predicate.go
JoshVanL: Thanks, I'll change this in a different PR to have separate concerns since the function itself wasn't added in this PR.

pkg/internal/approver/manager/predicate/predicate.go
JoshVanL: Thanks, added some code comments to make it clearer.

pkg/internal/approver/manager/predicate/predicate.go
JoshVanL: This then gets returned as an error, and basically halts the entire evaluation of the request until it is resolved.

I've added validation for this in the webhook, so will now be caught as CertificateRequestPolicy admission time.

pkg/internal/approver/manager/predicate/predicate.go
JoshVanL: See https://github.com/cert-manager/approver-policy/pull/98

JoshVanL:
Thank you for another review :slightly_smiling_face: @munnerz PTAL

inteon:
@JoshVanL how does order of precedence work?
eg: if there is a rule that matches namespace names "application-*" and a rule for "application-001", does "application-001" take precedence over the more general rule?

JoshVanL:
@inteon everything in the namespace selector must match that namespace in order for it to be eligible for evaluation for that policy

pkg/internal/approver/manager/predicate/predicate.go
```diff
@@ -77,6 +86,74 @@ func SelectorIssuerRef(_ context.Context, cr *cmapi.CertificateRequest, policies
...
+// that have an `spec.selector.namespace` matching the `metadata.namespace` of
+// the request. SelectorNamespace will match with `namespace.matchNames` on
+// namespaces using wilcards "*". Empty selector is equivalent to "*" and will
+// match on any Namespace.
+func SelectorNamespace(lister client.Reader) Predicate {
+	return func(ctx context.Context, request *cmapi.CertificateRequest, policies []policyapi.CertificateRequestPolicy) ([]policyapi.CertificateRequestPolicy, error) {
+		var matchingPolicies []policyapi.CertificateRequestPolicy
+
+		var namespace corev1.Namespace
+		if err := lister.Get(ctx, client.ObjectKey{Name: request.Namespace}, &namespace); err != nil {
```
inteon, line 99: Would it be possible to only execute this Get() call when we are certain that we have to retrieve the labels of the namespace?

inteon: eg. if no Namespace selectors are != nil, we can skip this retrieval

JoshVanL: We could.. however note that we are using a lister here which is metadata caching Namespaces so should be cheap.

I've gone ahead and added lazy fetching anyway.

JoshVanL: PTAL

inteon:
I agree with the design & the changes, but didn't test the PR locally yet.
/lgtm
/hold

inteon:
I tried testing this locally, using the following yaml manifest:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: sandbox
---
apiVersion: policy.cert-manager.io/v1alpha1
metadata:
  name: example-com
spec:
  allowed:
    commonName:
      value: "example.com"
    dnsNames:
      values:
      - "example.com"
      - "*.example.com"
    usages:
    - "server auth"

  constraints:
    privateKey:
      algorithm: RSA
      minSize: 2048

  selector:
    issuerRef:
      name: selfsigned-issuer
      kind: ClusterIssuer
      group: cert-manager.io
---
kind: Role
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: cert-manager-policy:example-com
  namespace: sandbox
rules:
- apiGroups: ["policy.cert-manager.io"]
  resources: ["certificaterequestpolicies"]
  verbs: ["use"]
  resourceNames: ["example-com"]
---
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: cert-manager-policy:example-com
  namespace: sandbox
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: cert-manager-policy:example-com
subjects:
# Policy intended to be used with a Certificate resource, so cert-manager is
# the user creating CertificateRequest. Bind to the cert-manager
# ServiceAccount.
- kind: ServiceAccount
  name: cert-manager
  namespace: cert-manager
---
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: selfsigned-issuer
spec:
  selfSigned: {}
---
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: my-selfsigned-ca
  namespace: default
spec:
  commonName: "example.com"
  dnsNames:
    - "example.com"
    - "aaa.example.com"
  usages:
  - "server auth"
  secretName: secret-test
  issuerRef:
    name: selfsigned-issuer
    kind: ClusterIssuer
    group: cert-manager.io
```
I expected the request to be approved, but this was not the case. Could you help me determine what is wrong?

JoshVanL:
@inteon the role has been created in `sandbox` namespace for cert-manager to use that policy, but the request has been created in the `default` namespace.

inteon:
/unhold
Thank you @JoshVanL.
I think this addition is useful for mid-complexity setups:
- allows filtering namespaces based on labels
- allows filtering namespaces using pattern matching

However, in some cases using a namespaced RBAC `RoleBinding` might be a better solution.
Also, more complex setups might want to use a rules engine.

Note: the functionality in this PR overlaps with what is possible using RBAC `RoleBinding`s (+ might be conflicting).
We decided that it is still ok to add this change because it covers the additional use cases mentioned above.

PR #97, opened 2022-07-11
Moves helm from being downloaded from curl, to being installed via hack/tools

JoshVanL:


hack/tools/go.mod
```diff
@@ -6,59 +6,133 @@ require (
 	github.com/norwoodj/helm-docs v1.10.0
 	github.com/onsi/ginkgo v1.16.5
 	github.com/princjef/gomarkdoc v0.3.2
+	helm.sh/helm/v3 v3.9.0
```
munnerz, line 9: How come we want to do this? It seems like it's quite a bit more expensive to build helm from scratch than it is to curl it, and I'm slightly concerned about awkward dependency conflicts in future too (what with us depending on controller-tools, kind, code-generator _and_ helm in one go.mod).

JoshVanL: The intention is to have as much of our version management in one tooling (go modules), rather than split across multiple (go modules, Makefile vars + curl).

SgtCoDFish: I'm a pretty strong -1 on actively _changing_ from downloading binaries to building using `go build`. I wouldn't see much of a need to go out and change a Makefile which uses `go build` already, but I don't think we should actually change. IMO it's going backwards.

I don't see how it's any easier to bump a version number in a `go.mod` than it is to bump a version number in a Makefile - _especially_ if it runs the risk of dependency hell as @munnerz points out.

IMO the ideal situation is downloading prebuilt binaries and checking their hashes to ensure they've not been tampered with. Fewer things that can go wrong, far less CPU usage, far less chance of Go dependency hell in the future. Also much easier to add [retries](https://github.com/cert-manager/cert-manager/blob/aeae4b35fced192c9d7f8dc11dfcb7f746ff4e2f/make/tools.mk#L55-L61) with curl if something goes wrong or there's a network flake.

> The intention is to have as much of our version management in one tooling (go modules), rather than split across multiple (go modules, Makefile vars + curl).

I'd argue we have two sets of tooling in this PR; we have two `go.mod` files, one for the application and one for tools. I think it's totally reasonable to have tool dependencies versioned in the Makefile and application dependencies versioned in the `go.mod`.

JoshVanL: Hear where you are both coming from. Since you're both against the PR I'll close it.

I think nix could be a really good solution to all of our tooling needs (:tm:) across cert-manager. The massive downside is that 1. developers will need to install `nix`, and 2. the team would need to learn `nix`. I'll perhaps add this as an agenda item for a future meeting.

/close

PR #119, opened 2022-09-13
Fire event on CertificateRequest resource _after_ status has been updated.

JoshVanL:
When the certificaterequest controller is about to update the status of the request with an approved or denied condition, it will fire an event. This event will be created _before_ the CertificateRequest is actually updated.

This can lead to inconsistent messaging in the cases where the update failed for one reason or another (typically RBAC related), but the request's events say that the request has been approved/denied.

This PR changes the order so that the event is created _after_ the update occurs.

---

note: The `ResultUnprocessed` and `default` cases have not been changed because they return early and don't actually update the resource.


irbekrm (approved):
Thanks Josh, this makes sense and is also how we do it elsewhere, i.e cert-manager.

/lgtm

PR #170, opened 2022-12-12
Add explicit namespaces to all namespaced resources in Helm charts

inteon:
This helps with templating, after running `helm template -n test ...` the namespaces are also included in the manifest.
This prevents the user from running `kubectl apply -n not-test -f rendered.yaml`. This is important because of references that still point to the `test` namespace.

### DIFF
--- a/deploy/charts/approver-policy/templates/deployment.yaml
+++ b/deploy/charts/approver-policy/templates/deployment.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ include "cert-manager-approver-policy.name" . }}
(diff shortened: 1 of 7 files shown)


SgtCoDFish (approved):
/lgtm
/approve

I can see that `Release.Namespace` is a [helm builtin](https://helm.sh/docs/chart_template_guide/builtin_objects/).

I ran `helm template -n abc123 deploy/charts/approver-policy > rendered.yaml`

And I can see that abc123 was referred to in the rendered YAML, but wasn't used. E.g. the `Deployment` has no namespace applied but has ` - --webhook-ca-secret-namespace=abc123` in its args.

Seems reasonable to add this to fix that edge case; thanks!

PR #186, opened 2023-01-23
Switch to SSA

inteon:
Switched both the `CertificateRequest` and `CertificateRequestPolicy` controllers to SSA.
This also required the tests to be updated (the fake client does not support mocking SSA).


irbekrm (approved):
Thanks @inteon I've read through the code- all makes sense to me and seems to be the same approach that we use elsewhere. This project does not have the same problem with moving to SSA as the other ones as we don't remove anything from CRPs/CRs.

I've played with the code a bit upgrading from an older version of approver-policy to this PR and checking that existing resources get correctly updated - all seems to work as expected.

I'm adding an approve with a hold as this PR hasn't been around for too long - in case anyone else wants to take a look, but happy for it to get merged.

Additionally, I think we need to update approver-policy docs to state that the oldest supported k8s version will be 1.22

/hold

irbekrm:
Thanks for fixing the boilerplate check as well!

/lgtm
/unhold

pkg/internal/controllers/certificaterequests.go
```diff
@@ -142,12 +146,35 @@ func addCertificateRequestController(ctx context.Context, opts Options) error {
 // function will call the approver manager to evaluate whether a
 // CertificateRequest should be approved, denied, or left alone.
 func (c *certificaterequests) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	result, patch, resultErr := c.reconcileStatusPatch(ctx, req)
+	if patch != nil {
+		con, patch, err := ssa_client.GenerateCertificateRequestStatusPatch(req.Name, req.Namespace, patch)
+		if err != nil {
+			err = fmt.Errorf("failed to generate connection patch: %w", err)
```
wallrj, line 153: Typo in error message.

Should say "failed to generate CertificateRequest.Status patch.

PR #184, opened 2023-01-23
Upgrade go dependencies

inteon:
Signed-off-by: Tim Ramlot <42113979+inteon@users.noreply.github.com>

PR #187, opened 2023-01-23
Upgrade hack/tools

inteon:
I missed these dependencies in https://github.com/cert-manager/approver-policy/pull/184.


SgtCoDFish (approved):
/lgtm
/approve

Thanks!

PR #189, opened 2023-01-24
Remove leader election in integration tests

inteon:
We can enforce tests to be run one after another without using leader election (setting procs=1 in ginkgo).
This PR also adds a check in code to make sure no two integration tests are running at the same time.
This removes the large amount of console messages related to leader election.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,5 +23,7 @@ bin
 *.swo
 *~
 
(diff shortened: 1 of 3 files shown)


SgtCoDFish (approved):
/lgtm
/approve

Looks like an improvement to me!

PR #190, opened 2023-01-24
BUGFIX & Run goroutine using manager

inteon:
Instead of spinning up a simple goroutine, we probably want to register the goroutine with the controller runtime manager. This will make sure that the goroutine is awaited before shutting down the controller manager.

**BUGFIX**: this PR also fixes a bug, after receiving the first value from the list of channels, the goroutine is stopped currently. Instead, we want to keep listening for additional events!


irbekrm:
Thanks @inteon, the bugfix is a good catch. The bug has been introduced in https://github.com/cert-manager/approver-policy/pull/39/commits/75f4c262c878c30e26ea0fe8cbbb50131d3f3598 by removing an enclosing function, but leaving the return statements.
Managing the goroutine like a c/r runnable also makes perfect sense and should make things more reliable.

/lgtm
/approve

PR #201, opened 2023-02-13
Improve makefile

inteon:
1. removes tests from 'verify' target
2. now, tests are run by the 'test' target
3. removed code-generation shell scripts that were not working properly for me locally
4. added better & more verify targets (like Helm lint & Kyverno pod best-practices)
5. Fixed Helm lint and Kyverno issues

Depends on https://github.com/jetstack/testing/pull/792


deploy/charts/approver-policy/README.md
```diff
@@ -1,6 +1,6 @@
 # cert-manager-approver-policy
 
-![Version: v0.4.2](https://img.shields.io/badge/Version-v0.4.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.4.0](https://img.shields.io/badge/AppVersion-v0.4.0-informational?style=flat-square)
+![Version: v0.0.0](https://img.shields.io/badge/Version-v0.0.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.0.0](https://img.shields.io/badge/AppVersion-v0.0.0-informational?style=flat-square)
```
inteon, line 3: The helm-docs tool uses the static Chart.yaml values, when we generate the .tar.gz chart release the versions are set correctly. I think it is acceptable to keep these versions in the README as v0.0.0.

deploy/charts/approver-policy/Chart.yaml
```diff
@@ -12,5 +12,5 @@ maintainers:
 sources:
 - https://github.com/cert-manager/approver-policy
 
-appVersion: v0.4.0
-version: v0.4.2
+appVersion: v0.0.0
+version: v0.0.0
```
inteon, line 16: I don't want to bother updating these versions anymore, I also updated the release logic so there is just 1 version (same version for chart and application).

deploy/charts/approver-policy/templates/deployment.yaml
```diff
@@ -51,18 +55,28 @@ spec:
...
+          name: temp-dir
 
         resources:
           {{- toYaml .Values.resources | nindent 10 }}
 
-      {{- if .Values.volumes }}
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities: { drop: ["ALL"] }
+          readOnlyRootFilesystem: true
```
maelvls, line 71: These improvements seem to be unrelated to this PR. Not a big deal though.

maelvls (approved):
I have not tested the release process, I will let you try it out.

Thanks for making this Makefile look like a bit more the other Makefiles we have in other projects.

One comment I had while reviewing is that the `securityContext` addition is slightly "out of focus" for this PR. But I don't think it is a big deal, and I suggest we merge this PR as-is.

deploy/charts/approver-policy/templates/deployment.yaml
inteon: I had to add these improvements in order for the kyverno test to not fail.

PR #208, opened 2023-03-04
Templating in policy

### LINKED ISSUE #62: CertificateRequestPolicy based on which namespace the certificate request belongs to

smuda:
This PR enables using templating in Policy with data from the request when validating the request.

One use case would be to allow dnsNames with the pattern "*.{{ .Request.Namespace }}.svc" but there are certainly other use cases.

This PR does not yet include any documentation (only comments in code) but if you agree this is a good idea, I'll add some as well.

Closes #62

### DIFF
--- a/pkg/internal/approver/allowed/evaluator.go
+++ b/pkg/internal/approver/allowed/evaluator.go
@@ -55,10 +55,12 @@ func (a allowed) Evaluate(_ context.Context, policy *policyapi.CertificateReques
 		return approver.EvaluationResponse{}, err
 	}
 
+	td := util.TemplateData{Request: request}
+
 	if len(csr.Subject.CommonName) > 0 {
 		if allowed.CommonName == nil || allowed.CommonName.Value == nil {
 			el = append(el, field.Invalid(fldPath.Child("commonName", "value"), csr.Subject.CommonName, "nil"))
-		} else if !util.WildcardMatches(*allowed.CommonName.Value, csr.Subject.CommonName) {
+		} else if !util.WildcardMatches(util.TemplateStr(td, *allowed.CommonName.Value), csr.Subject.CommonName) {
 			el = append(el, field.Invalid(fldPath.Child("commonName", "value"), csr.Subject.CommonName, *allowed.CommonName.Value))
 		}
 	} else if allowed.CommonName != nil && allowed.CommonName.Required != nil && *allowed.CommonName.Required {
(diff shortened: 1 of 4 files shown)


smuda:
I assume that you'll want me to update the website docs as well at [approver-policy.md](https://github.com/cert-manager/website/blob/master/content/docs/projects/approver-policy.md) but I'm going to wait for positive feedback before putting in any more work.

smuda:
@inteon No problem! You'll look at it when you have the time. Thank you for enabling tests.

inteon:
@smuda I'm wondering if CEL wouldn't be a better match for this kind of problem?

sastorsl:
Could you list a couple of `CertificateRequestPolicy` examples?
For example, are my use cases in #62 met by this PR?

smuda:
This is how one `CertificateRequestPolicy` would have looked with this PR:

```
apiVersion: policy.cert-manager.io/v1alpha1
metadata:
  name: my-policy
spec:
  allowed:
    dnsNames:
      values:
      - "*.{{ .Request.Namespace }}.svc"
```

This would have handled my usecase and I think it should have handled yours as well.

But it doesn't really matter since this PR was not accepted by the cert-manager team and after a discussion I'll go down a different route with another PR.

inteon:
> This is how one `CertificateRequestPolicy` would have looked with this PR:
> 
> ```
> apiVersion: policy.cert-manager.io/v1alpha1
> kind: CertificateRequestPolicy
> metadata:
>   name: my-policy
> spec:
>   allowed:
>     dnsNames:
>       values:
>       - "*.{{ .Request.Namespace }}.svc"
> ```
> 
> This would have handled my usecase and I think it should have handled yours as well.
> 
> But it doesn't really matter since this PR was not accepted by the cert-manager team and after a discussion I'll go down a different route with another PR.

Indeed, we did conclude that adding a full templating engine to this project opens up many new possible bugs and vulnerabilities. We agreed that it would be better to implement this functionality in a plugin first before adding it to this repo.

erikgb:
> We agreed that it would be better to implement this functionality in a plugin first before adding it to this repo.

Did someone start on a plugin? If not, I would like to take a shot at implementing a POC plugin based on CEL expressions declared as plugin values. With support for a limited set of CSR attributes. In our use case, we are interested in using this feature to validate `dnsNames` and `uris`, so I'll probably start with those.

Issue #213, opened 2023-03-17
Helm chart rendering error: converting YAML to JSON: yaml: line 61: did not find expected key


wallrj:
```yaml
volumeMounts:
  - name: ca-cert-example-volume
    mountPath: "/etc/ssl/certs/ca-cert-example-ca.crt"
    subPath: ca.crt
    readOnly: true
volumes:
  - name: ca-cert-example-volume
    configMap:
      name: ca-cert-example
      optional: false
```

```bash
$ helm template cert-manager-approver-policy --repo https://charts.jetstack.io  --version v0.6.2 --values values.yaml
Error: YAML parse error on cert-manager-approver-policy/templates/deployment.yaml: error converting YAML to JSON: yaml: line 61: did not find expected key

Use --debug flag to render out invalid YAML
```

Probably caused by incorrect indentation of the included yaml, here:

https://github.com/cert-manager/approver-policy/blob/19d9d8489a239ea74f11fe9da2f6d4a8b46b02d9/deploy/charts/approver-policy/templates/deployment.yaml#L62-L67

Introduced in:
 * https://github.com/cert-manager/approver-policy/pull/201

/kind bug

PR #214, opened 2023-03-17
Fix Helm chart volume mount indentation

wallrj:
Fixes: #213 

* Added a test to verify that Helm charts can be rendered
  And that values which are expected to be overridden, do not cause rendering errors.
  For example: https://prow.build-infra.jetstack.net/view/gs/jetstack-logs/pr-logs/pull/cert-manager_approver-policy/214/pull-cert-manager-approver-policy-verify/1636784667623428096#1:build-log.txt%3A165-171


inteon:
Thank you for fixing this issue.
/lgtm
/approve

PR #229, opened 2023-04-26
feat: fix app label of metrics svc for ServiceMonitor discovery

leotomas837:
The ServiceMonitor targets both the web hook service and the metrics service. Yet, only the metrics service must be scraped (the probe could be scraped too via the blackbox-exporter, but that is a subject for another PR). The webhook service must not be discovered by Prometheus.

### Prerequisites
Having a Kubernetes cluster running with the cert-manager/approver-policy chart installed and a prometheus instance ready to scrape metrics.

### How to reproduce

1. Make your Prometheus instance is scrape your approver-policy metrics. For this purpose, set the following values:
```yaml
nameOverride: cert-manager-approver-policy

app:
  metrics:
    port: 9402
    service:
      enabled: true
      servicemonitor:
        enabled: true
```

2. Open the Prometheus UI, you should see something similar:

Screenshot_2023-04-28_at_06_11_51-3-2

I have 2 replicas, hence the (4/4 up) in the target, but the point is that for a given replica, there is currently 2 services scraped: the webhook Service and the metrics Service. The metrics service should obviously be scraped, but not the webhook service (not even listening on port `9402` but on port `443`) as it is not serving any metrics. 

That is happening because the [ServiceMonitor here](https://github.com/cert-manager/approver-policy/blob/main/deploy/charts/approver-policy/templates/metrics-servicemonitor.yaml#L18) is scraping on the following app label:

```yaml
spec:
  selector:
    matchLabels:
      app: {{ include "cert-manager-approver-policy.name" . }}
```

Yet, both the metrics Service and the webhook Service have this label, check [here](https://github.com/cert-manager/approver-policy/blob/main/deploy/charts/approver-policy/templates/metrics-service.yaml#L8) and [here](https://github.com/cert-manager/approver-policy/blob/main/deploy/charts/approver-policy/templates/webhook.yaml#L7).

### How to test

By changing the `app` label of the metrics Service and setting the ServiceMonitor to select the new same `app` label, only the metrics Service gets scraped and you should see the following in your Prometheus UI (2/2 up) for 2 replicas:

Screenshot_2023-04-28_at_06_48_13

Simply apply the following JSON Patches to your chart installed in namespace `cert-manager` (I am using Helmfile so this is Helmfile syntax, but any tool applying the following JSON Patches obviously works):

```yaml
jsonPatches:
  # The ServiceMonitor currently targets both the metrics endpoint and the webhook endpoint
  - target:
      group: ""
      version: v1
      kind: Service
      name: cert-manager-approver-policy-metrics
      namespace: cert-manager
    patch:
      - op: replace
        path: "/metadata/labels/app"
        value: cert-manager-approver-policy-metrics
  - target:
      group: monitoring.coreos.com
      version: v1
      kind: ServiceMonitor
      name: cert-manager-approver-policy
      namespace: cert-manager
    patch:
      - op: replace
        path: "/spec/selector/matchLabels/app"
        value: cert-manager-approver-policy-metrics
```

### DIFF
--- a/deploy/charts/approver-policy/templates/metrics-service.yaml
+++ b/deploy/charts/approver-policy/templates/metrics-service.yaml
@@ -5,7 +5,7 @@ metadata:
   name: {{ include "cert-manager-approver-policy.name" . }}-metrics
   namespace: {{ .Release.Namespace | quote }}
   labels:
-    app: {{ include "cert-manager-approver-policy.name" . }}
+    app: {{ include "cert-manager-approver-policy.name" . }}-metrics
 {{ include "cert-manager-approver-policy.labels" . | indent 4 }}
 spec:
   type: {{ .Values.app.metrics.service.type }}
--- a/deploy/charts/approver-policy/templates/metrics-servicemonitor.yaml
+++ b/deploy/charts/approver-policy/templates/metrics-servicemonitor.yaml
@@ -15,7 +15,7 @@ spec:
   jobLabel: {{ include "cert-manager-approver-policy.name" . }}
   selector:
     matchLabels:
-      app: {{ include "cert-manager-approver-policy.name" . }}
+      app: {{ include "cert-manager-approver-policy.name" . }}-metrics
   namespaceSelector:
     matchNames:
       - {{ .Release.Namespace }}


wallrj (commented):
Thanks @leotomas837 

This looks ok to me, but could you add some instructions to the PR description explaining how to recreate the problem and how to test this improvement.

/ok-to-test

leotomas837:
@wallrj 

Instructions added

leotomas837:
@wallrj 

Did you get the chance to have a look on this ? I should be straight forward. Let me know if you need anything else from me.

wallrj:
More discussion here about confusing behaviour of `targetPort` : 
 * https://github.com/prometheus-operator/prometheus-operator/issues/2515

deploy/charts/approver-policy/templates/metrics-service.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: {{ include "cert-manager-approver-policy.name" . }}-metrics
   namespace: {{ .Release.Namespace | quote }}
   labels:
-    app: {{ include "cert-manager-approver-policy.name" . }}
+    app: {{ include "cert-manager-approver-policy.name" . }}-metrics
```
wallrj, line 8: The disadvantage of this, is that I can no longer easily discover all the resources related to this `app`. E.g.

```bash
$ kubectl get all --namespace cert-manager --selector app=cer
t-manager-approver-policy
NAME                                               READY   STATUS    RESTARTS   AGE
pod/cert-manager-approver-policy-c68dc44c9-5785f   1/1     Running   0          107m

NAME                                           TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
service/cert-manager-approver-policy           ClusterIP   10.96.10.146    <none>        443/TCP    107m
service/cert-manager-approver-policy-metrics   ClusterIP   10.96.133.139   <none>        9402/TCP   107m

NAME                                                     DESIRED   CURRENT   READY   AGE
replicaset.apps/cert-manager-approver-policy-c68dc44c9   1         1         1       107m
```

wallrj (changes requested):
I suggest we change `targetPort` to `port: metrics`, which has the same effect.
It seems to allow prometheus operator servicemonitor controller to select only the services that refer to ports with the name `metrics`.

leotomas837:
@wallrj 

I was looking at the same GitHub issue, and indeed you are right: `targetPort` setup the Prometheus's scrape config to scrape the container directly (for each service discovered, hence the duplicated target).
So 2 solutions: tightening the service discovery (my first approach), or targeting the service (by name) directly and not the container (your approach).

Happy to implement your approach, I like to think of a `ServiceMonitor` to actually be setup with service attributes and not container ones...

deploy/charts/approver-policy/templates/metrics-service.yaml
leotomas837: @wallrj

I was looking at the same GitHub issue, and indeed you are right: targetPort setup the Prometheus's scrape config to scrape the container directly (for each service discovered, hence the duplicated target).
So 2 solutions: tightening the service discovery (my first approach), or targeting the service (by name) directly and not the container (your approach).

Happy to implement your approach, I like to think of a ServiceMonitor to actually be setup with service attributes and not container ones...

wallrj: Great. Then please modify the PR to use the the `port: metrics` approach and ping me for another review.

wallrj:
Supplanted by #471

PR #233, opened 2023-05-10
Update dependencies, make changes to work with latest c/r

irbekrm:
This PR does the required work to bump dependencies, notably bumps kube libraries, cert-manager and controller-runtime.
The size of the PR is to do with the changes in controller-runtime as some functionality got removed or changed in a way that we needed to make a large number of changes to be able to bump c/r

⚠️ BREAKING ⚠️ path of the webhook validation handler has been changed from `/validate` to `/validate-policy-cert-manager-io-v1alpha1-certificaterequestpolicy`, see https://github.com/cert-manager/approver-policy/pull/233#discussion_r1194138573
This might cause a brief webhook downtime during upgrade where, if the path the webhook handler serves does not match the path in `ValidatingWebhookConfiguration`, it will not be possible to create or update `CertificateRequestPolicy` objects. Folks should verify that after upgrading to this version of approver-policy, `CertificateRequestPolicy`s can be applied as expected.

Changes:

- bumps to latest cert-manager version
- bump kube library deps and other deps to match cert-manager imports
- bump Go to 1.20 because kube 1.27 uses some functionality that was added in Go 1.20
- bumps deps of test tools
- updates builder's watch setup to fix issues related to changes in controller-runtime in https://github.com/kubernetes-sigs/controller-runtime/pull/2135/commits/dc62ed5dadc6ea2d3e22546368b939a2db1edfb2
- uses kustomize v4 -> v5 -needed because v4 is incompatible with the version of kustomize api module that is depended on by k8s.io/cli-runtime module that we use
- changes webhook setup- this is a big change to adopt the new webhook setup model from c/r see https://github.com/kubernetes-sigs/controller-runtime/blob/v0.15.0-alpha.0/examples/builtins/main.go#L74-L81. This change is needed in this PR because the latest c/r also removed webhook decoder injection functionality that we depend on in the current code

- updates base image


irbekrm:
The failures are to do with changes in controller-runtime

pkg/internal/controllers/certificaterequests.go
```diff
@@ -122,19 +121,19 @@ func addCertificateRequestController(ctx context.Context, opts Options) error {
...
 		// appropriate for a CertificateRequest. On RBAC events, Reconcile all
 		// CertificateRequests that are neither Approved or Denied.
 		// Only need to cache metadata for RBAC resources since we do not need any
 		// information in the spec.
-		Watches(&source.Kind{Type: new(rbacv1.Role)}, handler.EnqueueRequestsFromMapFunc(enqueueRequestFromMapFunc), builder.OnlyMetadata).
-		Watches(&source.Kind{Type: new(rbacv1.RoleBinding)}, handler.EnqueueRequestsFromMapFunc(enqueueRequestFromMapFunc), builder.OnlyMetadata).
-		Watches(&source.Kind{Type: new(rbacv1.ClusterRole)}, handler.EnqueueRequestsFromMapFunc(enqueueRequestFromMapFunc), builder.OnlyMetadata).
-		Watches(&source.Kind{Type: new(rbacv1.ClusterRoleBinding)}, handler.EnqueueRequestsFromMapFunc(enqueueRequestFromMapFunc), builder.OnlyMetadata).
-		Watches(&source.Kind{Type: new(corev1.Namespace)}, handler.EnqueueRequestsFromMapFunc(enqueueRequestFromMapFunc), builder.OnlyMetadata).
+		WatchesMetadata(new(rbacv1.Role), handler.EnqueueRequestsFromMapFunc(enqueueRequestFromMapFunc)).
```
irbekrm, line 132: This new watch type appears to be a wrapper that automatically appends the `OnlyMetadata` option https://github.com/kubernetes-sigs/controller-runtime/pull/2135/commits/dc62ed5dadc6ea2d3e22546368b939a2db1edfb2#diff-e77b9468ab935d5ea4d5cdae3b994114bada17df0001570e4b1436419afd50ccR166-R172

pkg/internal/controllers/certificaterequestpolicies.go
```diff
@@ -118,8 +118,8 @@ func addCertificateRequestPolicyController(_ context.Context, opts Options) erro
 
 	return ctrl.NewControllerManagedBy(opts.Manager).
 		For(new(policyapi.CertificateRequestPolicy)).
-		Watches(&source.Channel{Source: genericChan}, handler.EnqueueRequestsFromMapFunc(
-			func(obj client.Object) []reconcile.Request {
+		WatchesRawSource(&source.Channel{Source: genericChan}, handler.EnqueueRequestsFromMapFunc(
```
irbekrm, line 121: Since `Watches` now accepts `client.Object` only we here need to use the new lower level `WatchesRawSource` that has the same signature as `Watches` had up till now https://github.com/kubernetes-sigs/controller-runtime/pull/2135/commits/dc62ed5dadc6ea2d3e22546368b939a2db1edfb2#diff-e77b9468ab935d5ea4d5cdae3b994114bada17df0001570e4b1436419afd50ccR176-R179

pkg/internal/controllers/certificaterequests.go
```diff
@@ -122,19 +121,19 @@ func addCertificateRequestController(ctx context.Context, opts Options) error {
 		// Watch CertificateRequestPolicies. If a policy is created or updated,
 		// then we need to process all CertificateRequests that do not yet have an
 		// approved or denied condition since they may be relevant for the policy.
-		Watches(&source.Kind{Type: new(policyapi.CertificateRequestPolicy)}, handler.EnqueueRequestsFromMapFunc(enqueueRequestFromMapFunc)).
+		Watches(&policyapi.CertificateRequestPolicy{}, handler.EnqueueRequestsFromMapFunc(enqueueRequestFromMapFunc)).
```
irbekrm, line 124: `Watches` now expects `client.Object` instead of `source.Source`.
This change matches how c/r calls `Watches` in tests https://github.com/kubernetes-sigs/controller-runtime/pull/2135/commits/dc62ed5dadc6ea2d3e22546368b939a2db1edfb2#diff-64658a075291eb3beb906fd3850a3477413c70ed6bc8fb02590708ded0d6f3baR499

pkg/internal/webhook/validator_test.go
```diff
@@ -472,10 +472,7 @@ func Test_validatorHandle(t *testing.T) {
 				WithScheme(policyapi.GlobalScheme).
 				Build()
 
-			decoder, err := admission.NewDecoder(policyapi.GlobalScheme)
-			if err != nil {
-				t.Fatal(err)
-			}
+			decoder := admission.NewDecoder(policyapi.GlobalScheme)
```
irbekrm, line 475: `NewDecoder` now panics instead of returning an error https://github.com/kubernetes-sigs/controller-runtime/pull/2134/commits/927c09e54313d6476b5780b7f029f381d777c149#diff-80f5a051498d0c9b056952ccabe9e604c28b4061673003cd152d46581b25e785R35-R39

irbekrm:
Remaining errors are to do with health check being broken.

deploy/charts/approver-policy/templates/webhook.yaml
```diff
@@ -49,7 +49,7 @@ webhooks:
       service:
         name: {{ include "cert-manager-approver-policy.name" . }}
         namespace: {{ .Release.Namespace | quote }}
-        path: /validate
+        path: /validate-policy-cert-manager-io-v1alpha1-certificaterequestpolicy
```
irbekrm, line 52: How this path is generated seems to now be hardcoded in c/r https://github.com/kubernetes-sigs/controller-runtime/blob/v0.15.0-alpha.0/pkg/builder/webhook.go#L247-L250

pkg/internal/webhook/validator.go
```diff
@@ -43,59 +41,37 @@ type validator struct {
...
+func (v *validator) validate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	policy, ok := obj.(*policyapi.CertificateRequestPolicy)
+	if !ok {
+		return nil, fmt.Errorf("expected a CertificateRequestPolicy, but got a %T", obj)
+	}
 	var (
-		el      field.ErrorList
-		fldPath = field.NewPath("spec")
+		el field.ErrorList
+		// TODO: currently there will never be any warnings. Add
```
irbekrm, line 70: https://github.com/cert-manager/approver-policy/pull/235

irbekrm:
I can't reproduce this locally

/retest

deploy/charts/approver-policy/templates/webhook.yaml
munnerz: Hmm.. don't think it's a major issue/something we can really address easily, but this will create a breaking change at runtime too (requests will fail after this Webhook resource is updated, until the new version of the Deployment rolls out 😬)

irbekrm: Yes, I agree. I think in majority of cases folks should not notice anything, but I will add a clear warning to the release notes.

I don't see a straightforward way how to keep the existing path unless we don't use their new webhook builder mechanism and I think that writing a standalone webhook just to prevent this path change would be an overkill and too much of a maintenance burden

pkg/internal/webhook/validator.go
```diff
@@ -43,59 +41,37 @@ type validator struct {
...
-		return admission.Allowed("CertificateRequestPolicy validated")
+func (v *validator) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) (admission.Warnings, error) {
+	return v.validate(ctx, newObj)
+}
 
-	default:
-		return admission.Denied(fmt.Sprintf("validation request for unrecognised resource type: %s/%s %s", req.RequestKind.Group, req.RequestKind.Version, req.RequestKind.Kind))
-	}
+func (v *validator) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	return v.validate(ctx, obj)
```
inteon, line 58: @irbekrm why do we validate on delete?

inteon (changes requested):
LGTM, just one question: https://github.com/cert-manager/approver-policy/pull/233#discussion_r1207014363

pkg/internal/webhook/validator.go
irbekrm: 👍🏼 agree I've changed it in https://github.com/cert-manager/approver-policy/pull/233/commits/9bf645b2fda8506ea1478f8e188acfe98596c770

inteon (approved):
LGTM
/approve
/lgtm

PR #237, opened 2023-05-15
Bugfix: catch CRP not allowed cornercase

irbekrm:
This PR is based on #233 and should only be merged after #233 merges

Currently if an approver-policy plugin's validator evaluates that a `CertificateRequestPolicy is not allowed, but does not specify any errors for it, the `CertificateRequestPolicy` will be allowed, see [here]()

This PR ensures that any not-allowed `CertificateRequestPolicy` gets rejected by the validating webhook.


pkg/internal/webhook/validator.go
```diff
@@ -115,9 +116,17 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object) (admission
 			return nil, err
 		}
 		if !response.Allowed {
```
inteon, line 118:
```suggestion
		if !response.Allowed && len(response.Errors) == 0 {
		    el = append(el, errors.New("a plugin did not allow the CertificateRequest for unknown reasons"))
		}
		for _, e := range response.Errors {
		    el = append(el, e)
		}
```

Could you update the code as suggested, this would make it easier to understand for me.

irbekrm: I think that we should not use field error here to avoid misleading users suggesting that there is something wrong with the spec if the error is about something different, i.e some internal error etc.

inteon: Thanks for explaining, both comments make sense to me.

pkg/internal/webhook/validator_test.go
```diff
@@ -134,6 +137,23 @@ func Test_validate(t *testing.T) {
...
+					Selector: policyapi.CertificateRequestPolicySelector{
+						Namespace: &policyapi.CertificateRequestPolicySelectorNamespace{
+							MatchLabels: map[string]string{"foo": "bar"},
+						},
+					},
+				},
+			},
+			registeredPlugins: []string{"foo", "bar"},
+			webhooks:          []approver.Webhook{passingWebhook, notAllowedWebhookNoDetail},
+			wantsErr:          true,
```
inteon, line 155: Since we are not comparing the error messages anymore, it is not clear whether this change caused the error messages to change. 😟

pkg/internal/webhook/validator.go
```diff
@@ -93,38 +92,46 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object) (admission
...
+
+			allAllowed = false
 		}
 		warnings = append(warnings, response.Warnings...)
 	}
 
-	return warnings, el.ToAggregate()
+	var errs []error
+
+	if aggregateError := fieldErrs.ToAggregate(); aggregateError != nil {
```
inteon, line 125: @irbekrm `ToAggregate()` also de-duplicates all the errors, this was missing in the original PR

pkg/internal/webhook/validator.go
```diff
@@ -93,38 +92,46 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object) (admission
...
+		errs = append(errs, aggregateError.Errors()...)
+	}
+
+	// do not allow a CertificateRequestPolicy if it was not
+	// allowed by a plugin that did not set any errors
+	// TODO: when webhooks implement Name() method, provide a plugin name
+	if !allAllowed && len(errs) == 0 {
```
inteon, line 132: I'm using the `allAllowed` to make sure this error is also not double.

pkg/internal/webhook/validator_test.go
```diff
@@ -56,14 +57,16 @@ func Test_validate(t *testing.T) {
...
+
+		expectedWarnings admission.Warnings
+		expectedError    *string
 	}{
 		"if the object being validated is not a CertificateRequestPolicy return an error": {
-			crp:      &corev1.Pod{},
-			wantsErr: true,
+			crp: &corev1.Pod{},
+
+			expectedError: pointer.String("expected a CertificateRequestPolicy, but got a *v1.Pod"),
```
inteon, line 69: I added the literal errors in these tests, this way we know that these values do not change in this PR.

inteon:
@irbekrm I added a commit to this PR that fixes some of my concerns (see comments), PTAL.

pkg/internal/webhook/validator.go
```diff
@@ -36,8 +36,7 @@ import (
 
 // validator validates against policy.cert-manager.io resources.
 type validator struct {
-	lock sync.RWMutex
```
irbekrm, line 38: +1

pkg/internal/webhook/validator.go
irbekrm: yeah the boolean approach is better 👍🏼

pkg/internal/webhook/validator_test.go
irbekrm: I am on the edge with comparing error strings, but here the errors can come from different sources, so makes sense 👍🏼

PR #256, opened 2023-07-27
CEL expressions approver design

erikgb:
This PR contains an initial design proposal allowing policy authors to use CEL expressions to express policy rules. If this is ever implemented, it should address https://github.com/cert-manager/approver-policy/issues/62.

Also includes an initial design template copied from [cert-manager](https://github.com/cert-manager/cert-manager/blob/master/design/template.md).

/cc @inteon @SgtCoDFish

### DIFF
--- a/design/20230726-cel-policy.md
+++ b/design/20230726-cel-policy.md
@@ -1,0 +1,459 @@
+<!--
+This template is adapted from Kubernetes Enchancements KEP template https://raw.githubusercontent.com/kubernetes/enhancements/a86942e8ba802d0035ec7d4a9c992f03bca7dce9/keps/NNNN-kep-template/README.md
+-->
+
+# CEL validation rules to approve or deny a CertificateRequest
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories-optional)
+    - [X.509 DNS SAN in namespace domain](#x509-dns-san-in-namespace-domain)
+    - [SPIFFE ID identifying namespace](#spiffe-id-identifying-namespace)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Supported Versions](#supported-versions)
+- [Production Readiness](#production-readiness)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [CEL expressions with semantics similar to values](#cel-expressions-with-semantics-similar-to-values)
+  - [CEL expressions with full decoded CSR variable](#cel-expressions-with-full-decoded-csr-variable)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+This checklist contains actions which must be completed before a PR implementing this design can be merged.
+
+
+- [x] This design doc has been discussed and approved
+- [ ] Test plan has been agreed upon and the tests implemented 
+- [ ] Feature gate status has been agreed upon (whether the new functionality will be placed behind a feature gate or not)
+- [ ] Graduation criteria is in place if required (if the new functionality is placed behind a feature gate, how will it graduate between stages)
+- [ ] User-facing documentation has been PR-ed against the release branch in [cert-manager/website]
+
+
+## Summary
+
+<!--
+This section is important for producing high-quality, user-focused
+documentation such as release notes.
+
+A good summary is probably around a paragraph in length.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+While approver-policy allows policies to include namespace selectors, this does not scale when attempting to
+enforce CSR attributes that must be a function of the CertificateRequest's namespace (or other CertificateRequest fields).
+This proposal suggests an extension to the CertificateRequestPolicy API allowing policy authors to specify approval rules using
+[CEL in Kubernetes](https://kubernetes.io/docs/reference/using-api/cel/) using variables with information from the CertificateRequest.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals, and non-goals of
+the proposed enhancement.  Describe why the change is important and the benefits to users. The
+motivation section can optionally provide links to
+demonstrate the interest in this functionality amongst the community.
+-->
+
+### Goals
+
+<!--
+List specific goals. What is this proposal trying to achieve? How will we
+know that this has succeeded?
+-->
+
+- allow policy authors to express `allowed` CSR attribute values to be a function of CertificateRequest's fields using CEL expressions
+- include variables with `namespace` and `name` of the CertificateRequest in the CEL context provided to expressions
+- validate CEL expressions in CertificateRequestPolicy on admission
+- extensions to the CertificateRequestPolicy API should be backwards compatible
+
+### Non-Goals
+
+<!--
+What is out of scope for this proposal? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+- CEL support in CertificateRequestPolicy `constraints`
+- CEL variable with decoded CSR in the CEL context provided to expressions
+
+## Proposal
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+What is the desired outcome and how do we measure success?
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation- those should go into "Design Details" below.
+-->
+
+The current CertificateRequestPolicy API allows a policy author to specify `allowed` CSR attributes in the
+CertificateRequest, and will deny the request if it contains an attribute which is not present in the `allowed` block.
+To configure valid CSR attribute values, the API provides fields for allowed
+`values` (array; any element matching the attribute is sufficient),
+`value` (must match the attribute) or an array of literals/enums (in `usages`) depending on the requested attribute.
+Allowed string fields accept wildcards `"*"` within their values.
+
+This proposal suggests adding support for specifying allowed string fields using CEL expressions, similar to the existing
+wildcard support. While CEL expressions could eventually replace the current CSR attribute validation API - including the 
+wildcard support, the main improvement would be support for policies that specify an allowed attribute value that depends
+on CertificateRequest fields, e.g. `namespace`. If implemented, this new feature would close
+[#62](https://github.com/cert-manager/approver-policy/issues/62).
+
+An [experimental approver-policy plugin](https://github.com/erikgb/cel-approver-policy-plugin) providing some of the
+requested features exists, and the proposal (including alternatives) is inspired by this plugin.
+The main downside of using a plugin for this, in addition to increased maintenance/complexity,
+is that the approver-policy rules are still in play.
+Which means that approver rules are somehow duplicated in the `allowed` block and in the plugin configuration but
+this should be avoided as maintaining things in multiple places is error-prone.
+
+
+### User Stories (Optional)
+
+<!--
+Detail the things that people will be able to do if this proposal gets implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### X.509 DNS SAN in namespace domain
+
+As a cluster admin in a multi-tenant cluster, I want to provide a cert-manager ClusterIssuer to our users allowing
+self-provisioning of server certificates. Since the cluster is multi-tenant, I want to ensure that tenants do not
+"trip on each other's toes" and enforce that the issuer only signs certificates where the X.509 DNS SAN attribute
+ends with `<namespace>.<domain>`.
+
+A variant of this user story is exemplified in [#62](https://github.com/cert-manager/approver-policy/issues/62),
+and I think it's one of the most common user stories the proposed feature will solve.
+
+#### SPIFFE ID identifying namespace
+
+As a cluster admin in a multi-tenant cluster, I want to provide a cert-manager ClusterIssuer to our users allowing
+self-provisioning of [X.509 SVID certificates](https://github.com/spiffe/spiffe/blob/main/standards/X509-SVID.md).
+The [SPIFFE ID](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE-ID.md) (X.509 URI SAN) should identify
+the namespace (and service account), and must start with `spiffe://<trust-domain>/ns/<namespace>/sa/`.
+
+### Notes/Constraints/Caveats (Optional)
+
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go into as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+N/A
+
+### Risks and Mitigations
+
+<!--
+What are the risks of this proposal, and how do we mitigate? Think broadly.
+For example, consider both security and how this will impact the larger
+Kubernetes/PKI ecosystem.
+
+-->
+
+N/A
+
+## Design Details
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+The proposal is to add sibling properties to `values` and `value` properties inside the `allowed` block.
+The new fields should contain CEL validations used to validate CSR attribute values with similar
+[behavior](https://cert-manager.io/docs/projects/approver-policy/#allowed) as approver-policy has today,
+but with the following extension: If validations are present for the given CSR attribute, and **ALL**
+validations pass for **ALL* attribute values, the attribute will be allowed.
+
+Note: To ensure we don't break behaviors in the existing API version, the following should be enforced:
+
+* if `required` is set to `true` in policy and attribute value empty in request, the request will be denied
+  by policy without running any validations.
+* if `required` is set to `true` in policy, the policy will need specify `values`/`value` or `validations`.
+  This should be enforced by the webhook.
+* if attribute value not empty in request, the request will be denied by policy if `values`/`value`
+  and `validations` are both unset.
+
+Most CSR attributes can have multiple values, and it could make sense to have validations on the list of
+attribute values. Example: Limit the number of `dnsNames`. But we think that the most common use case is to
+validate attribute value (per-item validation). Support for more complex validations, including cross-attribute
+validation, can eventually be added later.
+
+The validations themselves should be expressed using the same API as Kubernetes `x-kubernetes-validations`, having
+a mandatory `rule` field for the CEL expression and an optional `message` field for the validation error message.
+
+A request attribute value could be allowed by the policy if `values`/`value` or `validations` are specified.
+
+To use validations in CertificateRequestPolicy the API may look like this (example):
+
+```yaml
+spec:
+  allowed:
+    dnsNames:
+      validations:
+        - message: only dnsNames in the local namespace service domain is allowed.
+          rule: self.endsWith(cr.namespace + '.svc') || self.endsWith(cr.namespace + '.svc.cluster.local')
+```
+
+The CEL context variable `self` is inspired by
+[CEL expressions in Kubernetes](https://kubernetes.io/docs/reference/using-api/cel/),
+and should appear familiar to our users. 
+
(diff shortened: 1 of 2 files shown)


design/20230726-cel-policy.md
```diff
@@ -0,0 +1,368 @@
...
+expressions and is preferred for this reason (IMO). 
+
+It should be allowed to mix `values` and `expressions`, so the following policy should also be acceptable:
+
+```yaml
+spec:
+  allowed:
+    dnsNames:
+      expressions:
+        - "self.endsWith(cr.namespace + '.apps.my-cluster.com')"
```
erikgb, line 204:
```suggestion
      validations:
        - rule: "self.endsWith(cr.namespace + '.apps.my-cluster.com')"
          message: "DNSName should be in cluster apps domain"
```

After a bit of thinking, I think we should align our design closer to upstream [validation rules](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules).

I'll update the suggested design now.

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,368 @@
...
+you're proposing, but should not include things like API designs or
+implementation- those should go into "Design Details" below.
+-->
+
+The current CertificateRequestPolicy API allows a policy author to specify `allowed` CSR attributes in the
+CertificateRequest, and will deny the request if it contains an attribute which is not present in the `allowed` block.
+To configure valid CSR attribute values, the API provides fields for allowed
+`values` (array; any element matching the attribute is sufficient),
+`value` (must match the attribute) or an array of literals/enums (in `usages`) depending on the request attribute.
+Allowed string fields accept wildcards "*" within its value(s).
```
jsoref, line 103:
```suggestion
Allowed string fields accept wildcards `"*"` within its value(s).
```
? Or `*` ...

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,368 @@
...
+
+A good summary is probably around a paragraph in length.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+While approver-policy allows policies to include namespace selectors, this does not scale when attempting to
+enforce CSR attributes that must be a function of the CertificateRequest namespace (or other CertificateRequest fields).
+This proposal suggests an extension to the CertificateRequestPolicy API allowing policy authors to specify approval rules using
+[CEL expressions](https://github.com/google/cel-spec) containing variables with information from the CertificateRequest.
```
jsoref, line 56: Is there a reason to point to that instead of https://kubernetes.io/docs/reference/using-api/cel/?

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,368 @@
...
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### X.509 DNS SAN in namespace domain
+
+As a cluster admin in a multi-tenant cluster, I want to provide a cert-manager ClusterIssuer to our users allowing
+self-provisioning of server certificates. Since the cluster is multi-tenant, I want to ensure that tenants do not
+"tip on each other's toes" and enforce that the issuer only signs certificates where the X.509 DNS SAN attribute
```
jsoref, line 124:
```suggestion
"trip on each other's toes" and enforce that the issuer only signs certificates where the X.509 DNS SAN attribute
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,368 @@
...
+ends with `<namespace>.<domain>`.
+
+A variant of this user story is exemplified in [#62](https://github.com/cert-manager/approver-policy/issues/62),
+and I think it's one of the most common user stories the proposed feature will solve.
+
+#### SPIFFE ID identifying namespace
+
+As a cluster admin in a multi-tenant cluster, I want to provide a cert-manager ClusterIssuer to our users allowing
+self-provisioning of X.509 SVID certificates. The SPIFFE ID (X.509 URI SAN) should identify the namespace
```
jsoref, line 133: It'd be nice if `SVID` and `SPIFFE ID` were links to things.

design/20230726-cel-policy.md
erikgb: We could link to both. The Kubernetes docs give a good overview, but if you're interested in the details, I think you need to consult the detailed spec.

design/20230726-cel-policy.md
erikgb: Thanks @jsoref! Updated with links.

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,369 @@
...
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+An [experimental approver-policy plugin](https://github.com/erikgb/cel-approver-policy-plugin) providing some of the
+requested features exists, and it's proposed to base the design on this plugin.
```
jsoref, line 165: this line is awkward. https://en.wikipedia.org/wiki/Garden-path_sentence

In general, I'd discourage the use of abbreviations (`'`), 
```suggestion
requested features exists, and we propose to base the design on this plugin.
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,369 @@
...
+Pros:
+
+- Extremely flexible. I.e. allows for CSR cross-attribute validation.
+- Some would find this API more expressive and better describe the intentions.
+
+Cons:
+
+- More complex expressions that will require more in-depth knowledge of CEL.
+- Will create a "split" API, since you would probably use either `expressions` or `allowed`/`constraints`.
+  This could increase the maintenance burden and source to subtle bugs. Breaking the API could be an option to mend this.
```
jsoref, line 356:
```suggestion
  This could increase the maintenance burden and lead to subtle bugs. Breaking the API could be an option to mitigate this.
```

design/20230726-cel-policy.md
erikgb: Good point, thanks! I will change the link in my next round on this. Also thanks for the rewording suggestions! 👍

design/20230726-cel-policy.md
erikgb: Modified the proposal to reuse the Kubernetes CEL validations.

erikgb:
I have now updated the proposal and moved the previous proposal into the alternatives. The new proposal is now significantly closer to the upstream CEL validations rules API. Please take another look. I would love to see some progress on this design proposal so we eventually can move over to implementing it. 😸 

/cc @jsoref @inteon @SgtCoDFish @wallrj

SgtCoDFish (commented):
/lgtm
/approve
/hold

(Holding so others can chime in)

First of all, this is well-written and makes sense to me. I understand the rationale and thank you for your patience with this!

I think the idea here is solid. It adds a lot of flexibility and power and the tradeoff in complexity is worth it.

A big part of why I'm really relaxed about this is that approver-policy isn't super-widely used and we're still at v1alpha1, which to me means we should be more relaxed about making changes and trying things out.

I think we should proceed with this on the basis that we _should_ be experimenting with this. I'd also support making these changes in v1alpha2, leaving v1alpha1 as what we have today, which in my mind should make this even easier to approve. But I don't think that's required - I'd be happy to add this to v1alpha1.

As for the alternatives: I instinctively prefer having expressions be tied to fields rather than operating on the whole CSR (i.e. I prerfer the proposal to that alternative). I think the simpler the CEL the better.

inteon (approved):
+1

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
+<!--
+This template is adapted from Kubernetes Enchancements KEP template https://raw.githubusercontent.com/kubernetes/enhancements/a86942e8ba802d0035ec7d4a9c992f03bca7dce9/keps/NNNN-kep-template/README.md
+-->
+
+# CEL validation rules to approve or deny CertificateRequest
```
jsoref, line 5:
```suggestion
# CEL validation rules to approve or deny a CertificateRequest
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
...
+This section is important for producing high-quality, user-focused
+documentation such as release notes.
+
+A good summary is probably around a paragraph in length.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+While approver-policy allows policies to include namespace selectors, this does not scale when attempting to
+enforce CSR attributes that must be a function of the CertificateRequest namespace (or other CertificateRequest fields).
```
jsoref, line 55:
```suggestion
enforce CSR attributes that must be a function of the CertificateRequest's namespace (or other CertificateRequest fields).
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
...
+-->
+
+### Goals
+
+<!--
+List specific goals. What is this proposal trying to achieve? How will we
+know that this has succeeded?
+-->
+
+- allow policy authors to express `allowed` CSR attribute value(s) to be a function of CertificateRequest fields using CEL expressions
```
jsoref, line 75:
```suggestion
- allow policy authors to express `allowed` CSR attribute value(s) to be a function of CertificateRequest's fields using CEL expressions
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
...
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation- those should go into "Design Details" below.
+-->
+
+The current CertificateRequestPolicy API allows a policy author to specify `allowed` CSR attributes in the
+CertificateRequest, and will deny the request if it contains an attribute which is not present in the `allowed` block.
+To configure valid CSR attribute values, the API provides fields for allowed
+`values` (array; any element matching the attribute is sufficient),
+`value` (must match the attribute) or an array of literals/enums (in `usages`) depending on the request attribute.
```
jsoref, line 103: possibly?
```suggestion
`value` (must match the attribute) or an array of literals/enums (in `usages`) depending on the requested attribute.
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
...
+Allowed string fields accept wildcards `"*"` within its value(s).
+
+This proposal suggests to add support for specifying allowed string fields using CEL expressions, similar to the existing
+wildcard support. While CEL expressions could eventually replace the current CSR attribute validation API - including the 
+wildcard support, the main improvement would be support for policies that specifies an allowed attribute value that depends
```
jsoref, line 108:
```suggestion
wildcard support, the main improvement would be support for policies that specify an allowed attribute value that depends
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
...
+on CertificateRequest fields, e.g. `namespace`. If implemented, this new feature would close
+[#62](https://github.com/cert-manager/approver-policy/issues/62).
+
+An [experimental approver-policy plugin](https://github.com/erikgb/cel-approver-policy-plugin) providing some of the
+requested features exists, and the proposal (including alternatives) is inspired by this plugin.
+The main downside of using a plugin for this, in addition increased maintenance/complexity,
```
jsoref, line 114:
```suggestion
The main downside of using a plugin for this, in addition to increased maintenance/complexity,
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
...
+is that the approver-policy rules are still in play.
+Which means that approver rules are somehow duplicated in the `allowed` block and in the plugin configuration and
+should be avoided.
```
jsoref, line 117:
```suggestion
Which means that approver rules are somehow duplicated in the `allowed` block and in the plugin configuration but
this should be avoided as maintaining things in multiple places is error prone.
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
...
+
+## Alternatives
+
+<!--
+What other approaches did you consider, and why did you rule them out? These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+
+### CEL expressions with similar semantics as values
```
jsoref, line 333:
```suggestion
### CEL expressions with semantics similar to values
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
...
+
+_Note: This was the original proposal, after some consideration it is considered inferior to the current proposal,
```
jsoref, line 335:
```suggestion
_Note: This was the original proposal, after some consideration it was considered inferior to the current proposal,
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
...
+```
+
+### CEL expressions with full decoded CSR variable
+
+This alternative was suggested by @inteon on Slack.
+
+The idea here is to provide the policy author with a CEL context variable containing the decoded CSR replacing the
+`self` variable in the proposed design. Since the expressions now get access to the full CSR, the expressions will have
+to be "promoted" in the CertificateRequestPolicy API. The expression will have to return a bool indicating approved or
+denied. An example policy with this alternative API might look like (example);
```
jsoref, line 384:
```suggestion
denied. An example policy with this alternative API might look like (example):
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
...
+
+To use expressions in CertificateRequestPolicy the API may look like this (example):
+
+```yaml
+spec:
+  allowed:
+    dnsNames:
+      expressions:
+        - "self.endsWith(cr.namespace + '.svc')"
+        - "self.endsWith(cr.namespace + '.svc.cluster.local')"
```
inteon, line 357: Maybe, it would be useful if we could also express that eg. you can have at most 3 dnsNames in your cert?

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
...
+
+It should be allowed to mix `values` and `validations`, so the following policy should also be acceptable:
+
+```yaml
+spec:
+  allowed:
+    dnsNames:
+      validations:
+        - message: dnsName can be maximum 64 characters.
+          rule: self.size() =< 64
```
inteon, line 205: How could we limit the number of dnsNames? Could we eg. have a "selfs" parameter?

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,418 @@
...
+
+To use validations in CertificateRequestPolicy the API may look like this (example):
+
+```yaml
+spec:
+  allowed:
+    dnsNames:
+      validations:
+        - message: only dnsNames in the local namespace service domain is allowed.
+          rule: self.endsWith(cr.namespace + '.svc') || self.endsWith(cr.namespace + '.svc.cluster.local')
```
inteon, line 190: I would start with a `cr` value that only contains the basic fields of the CertifiicateRequest resource (name, namespace, annotations, labels, uuid, ....). Once that works, we could also add other properties (like x509 stuff).

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,369 @@
...
+Let me start with a quote from the
+[approver-policy documentation](https://cert-manager.io/docs/projects/approver-policy/#allowed):
+
+> Allowed is the block that defines attributes that match against the corresponding attribute in the request.
+> A request is permitted by the policy if the request omits an allowed attribute, but will deny the request if
+> it contains an attribute which is not present in the allowed block.
+
+The last sentence describes the requirement that could need custom CEL function(s) to allow the user to specify allowed
+CSR attributes. We cannot expect the user to include validations for **all** CSR attributes in the expression(s). So I
+think we at least need a function to list allowed CSR attribute usage.
```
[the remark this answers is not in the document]
erikgb: I have now looked at the source code, and approver-policy only evaluates a policy against the CSR attributes that approver-policy supports. So you are correct that it might be considered too permissive in this sense. I suggest creating an issue for this if it is something that we should attempt to fix. Changing it would be breaking tho. 😉

design/20230726-cel-policy.md
erikgb: This is an interesting and important comment! Strictly speaking, many CSR attributes are multi-valued - so it could make sense to make `self` refer to the list of attribute values in the CSR - which will allow for the validation you are requesting here. But I suspect this will make the feature less user-friendly. The validation example this review comment refers to would then have to be expressed as `self.all(dnsName, dnsName.size() =< 64)` - which is not too bad IMO.

But this could easily lead to user errors like `self.size() =< 64` would now limit the number of dnsNames in the CSR to 63. 😉 Which is probably not what the user had in mind.

design/20230726-cel-policy.md
erikgb: Agreed, I think we should start even simpler in the initial implementation and provide just CR `namespace` and `name`. I will add this initial MVP to the proposed design.

design/20230726-cel-policy.md
erikgb: IMO the most user-friendly is to keep the validations under `allowed`  as per-item validations. This is how the `value`/`values` fields in the current API work. More complex CEL validations, typically cross-cutting concerns could be introduced later outside of the `allowed` API structure. I am updating the proposal with this clarification now.

design/20230726-cel-policy.md
erikgb: Updated, PTAL!

SgtCoDFish (approved):
/lgtm
/approve

Fundamentally, I think we should be experimenting with approver-policy. Now's the time for it!

I'd be much more conservative about this if it were for cert-manager itself... but it's not. I think this is worth a go. It makes sense. It's well written. Thank you for this Erik! 👍

erikgb:
I have initiated the work on implementing this new feature: https://github.com/cert-manager/approver-policy/pull/277
Please take a look if you have time.

erikgb:
After working on improving the API docs and understanding more of the current API behavior, I have updated the proposed design (and examples with the following):
>Note: To ensure we don't break behaviors in the existing API version, any validations can never allow a request otherwise denied by the `values`/`value` and `required` fields.

I think that is our safest option. The next API version could make this feature even more user-friendly. But I think it will be acceptable as proposed now.

I am almost ready to move over to the implementation of this, so I would prefer a more "formal" approval of this design proposal by merging this PR. 😉 PTAL!

hawksight:
Just wanted to add that the design I read just over 2 weeks ago looked great. It covered a few things I had not thought of and there was nothing I disagreed with. Now work has started, I will attempt to look over for understanding.

The only minor thing is that there are a few empty sections in the design doc, might it be worth either a quick comment of `N/A` if the section isn't applicable.

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,459 @@
...
+you're proposing, but should not include things like API designs or
+implementation- those should go into "Design Details" below.
+-->
+
+The current CertificateRequestPolicy API allows a policy author to specify `allowed` CSR attributes in the
+CertificateRequest, and will deny the request if it contains an attribute which is not present in the `allowed` block.
+To configure valid CSR attribute values, the API provides fields for allowed
+`values` (array; any element matching the attribute is sufficient),
+`value` (must match the attribute) or an array of literals/enums (in `usages`) depending on the requested attribute.
+Allowed string fields accept wildcards `"*"` within its value(s).
```
jsoref, line 105:
```suggestion
Allowed string fields accept wildcards `"*"` within their values.
```
I think...

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,459 @@
...
+
+This proposal suggests to add support for specifying allowed string fields using CEL expressions, similar to the existing
```
jsoref, line 107:
```suggestion
This proposal suggests adding support for specifying allowed string fields using CEL expressions, similar to the existing
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,459 @@
...
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+The proposal is to add sibling properties to `values` and `value` properties inside the `allowed` block.
+The new fields should contain CEL validations used to validate CSR attribute values with similar
+[behavior](https://cert-manager.io/docs/projects/approver-policy/#allowed) as approver-policy has today,
+but with the following extension: If validations are present for the given CSR attribute, and **ALL**
+validations passes for **ALL* attribute values, the attribute will be allowed.
```
jsoref, line 182:
```suggestion
validations pass for **ALL* attribute values, the attribute will be allowed.
```

design/20230726-cel-policy.md
```diff
@@ -0,0 +1,459 @@
...
+
+Note: To ensure we don't break behaviors in the existing API version, the following should be enforced:
+
+* if `required` set to `true` in policy and attribute value empty in request, the request will be denied
+  by policy without running any validations.
+* if `required` set to `true` in policy, the policy will need specify `values`/`value` or `validations`.
```
jsoref, line 188:
```suggestion
* if `required` is set to `true` in policy and attribute value empty in request, the request will be denied
  by policy without running any validations.
* if `required` is set to `true` in policy, the policy will need specify `values`/`value` or `validations`.
```

PR #259, opened 2023-08-23
Remove the command field from the yaml deployment

inteon:
This PR makes it easier to swap the image with an image with another entrypoint (eg. because you are running approver-policy with additional plugins)


wallrj (approved):
Thanks @inteon 

The smoke tests passed.
I went and reminded myself that they do indeed generate the Helm chart, install it and then exercise the approver-policy controller.

Looking forward to being able to use `ko` built images (with their unusual entry-point path)

/lgtm

PR #265, opened 2023-09-18
Enable grouping for dependabot

inteon:
This PR adds grouping options for dependabot, it will make dependabot create single grouped PRs to upgrade deps.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

see https://github.com/cert-manager/issuer-lib/pull/65 & https://github.com/cert-manager/issuer-lib/pull/67 for working example

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -5,7 +5,13 @@ updates:
   directory: /
   schedule:
     interval: weekly
+  groups:
+    all:
+      patterns: ["*"]
 - package-ecosystem: github-actions
   directory: /
   schedule:
     interval: weekly
+  groups:
+    all:
+      patterns: ["*"]


erikgb (commented):
Do we really want to group **everything**? I think the cert-manager dependency will control most of the "problematic" dependencies. Could it be an option to ignore k8s dependencies (including controller-runtime) in Dependabot configuration?

inteon:
> Do we really want to group **everything**? I think the cert-manager dependency will control most of the "problematic" dependencies. Could it be an option to ignore k8s dependencies (including controller-runtime) in Dependabot configuration?

I think it is the easiest solution right now:
1. similarly to how I upgrade dependencies, it upgrades everything at the same time (we can add any necessary additional to this PR to make the PR pass tests)
2. bundling all upgrades in 1 PR greatly reduces the noise from all these dependabot PRs.

I think this is a very good start & if necessary, we could split the PRs into more groups.

erikgb:
SGTM! 🚀 🚢

maelvls:
/approve

Thank you @erikgb for reviewing!

PR #272, opened 2023-09-22
Refactor Evaluate preparing for CEL

erikgb:
When/if the CEL feature proposed in https://github.com/cert-manager/approver-policy/pull/256 is implemented, we need to change the policy evaluation logic. As the code is already quite complex and repetitive, I am in this PR suggesting a refactoring that will make the code easier to change, and review in follow-up PRs.

/cc @inteon @SgtCoDFish


inteon (approved):
Thanks, great to see this simplification!
/approve
/lgtm

PR #277, opened 2023-10-02
feat: add allowed CEL validation rules

### LINKED ISSUE #62: CertificateRequestPolicy based on which namespace the certificate request belongs to

### LINKED ISSUE #149: Regex to disallow wildcard certificates
Is it currently possible? From what I read, I thought it might but when I tested it did not work as I expected.

https://cert-manager.io/docs/projects/approver-policy/

```
spec:
  allowed:
    commonName:
      required: true
      value: '*'
    dnsNames:
      required: true
      values:
      - "\*.domain.com"
```

I would like a certificate request consisting of a dnsName with the suffix `domain.com` to be allowed, but not a wildcard itself `*.domain.com`.

```
  - lastTransitionTime: "2022-10-25T18:41:53Z"
    message: 'No policy approved this request: [default-policy: spec.allowed.dnsNames.values:
      Invalid value: []string{"test-cert.domain.com"}: 
      \*.domain.com]'
```

erikgb:
Implements the design proposed in https://github.com/cert-manager/approver-policy/pull/256.

Fixes #62, fixes #149


pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -178,24 +178,33 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 	// Accepts wildcards "*".
 	// An omitted field or value of `nil` forbids any value on the related field
-	// in the request from being requested.
-	// An empty slice `[]` is equivalent to `nil`, however an empty slice pared
-	// with Required `true` is an impossible condition that always denies.
-	// Values may not be `nil` if Required is `true`.
+	// in the request from being requested unless Validations are present.
+	// Values and Validations may not be both be empty if Required is `true`.
 	// +optional
 	Values *[]string `json:"values,omitempty"`
```
erikgb, line 190: I don't know why this field is defined as a **pointer** of string slice. IMO treating nil and empty slice differently in an API is an anti-pattern, but I presume there is a reason for this here? 🤔

inteon: It would be very useful to properly document the current behaviour. Imo the current documentation is not very easy to understand.

erikgb: > It would be very useful to properly document the current behaviour.

I agree with this - even if the behavior is well documented on our website. @inteon, are you suggesting a pre-PR to brush up on the current Go docs? IMO this should be written with the Kubernetes API in mind. As it stands now, it's more a documentation of the Go API.

inteon: If you could create such a pre-PR this, that would be awesome!

erikgb: https://github.com/cert-manager/approver-policy/pull/280

erikgb:
@inteon @SgtCoDFish Do you have time for an early review of the work so far? I need to:

- implement the validator (which I will "steal" from my cel-plugin)
- add compilation of CEL in the webhook call path to ensure it's valid on policy admission
- add tests

But apart from such "minor" details 😉, it shows how I am planning to introduce opt-in validations to the evaluator. IMO using init-functions to wire up code is generally an anti-pattern and should be avoided. Should I create another pre-PR to fix this? If the evaluator (`allowed`) is wired up "the normal way", it is much easier to create a design that allows compiled CEL programs to be cached.

inteon:
> @inteon @SgtCoDFish Do you have time for an early review of the work so far? I still need to:
> 
> * implement the validator (which I will "steal" from my cel-plugin)
> * add compilation of CEL in the webhook call path to ensure it's valid on policy admission
> * add tests
> 
> But apart from such "minor" details 😉, it shows how I am planning to introduce opt-in validations to the evaluator. IMO using init-functions to wire up code is generally an anti-pattern and should be avoided. Should I create another pre-PR to fix this? If the evaluator (`allowed`) is wired up "the normal way", it is much easier to create a design that allows compiled CEL programs to be cached.

Looks good, what do you think about caching those CEL rules? Would that be hard to implement?

erikgb:
> Looks good, what do you think about caching those CEL rules? Would that be hard to implement?

With the current wiring in init-funcs it must be an "internal"/hidden cache. But yes, it should be possible to do something simple to avoid compiling all the time. Do we have any preference for caching implementation?

erikgb:
> > Looks good, what do you think about caching those CEL rules? Would that be hard to implement?
> 
> With the current wiring in init-funcs it must be an "internal"/hidden cache. But yes, it should be possible to do something simple to avoid compiling all the time. Do we have any preference for caching implementation?

The caching should now work. It's a pretty simple cache at the moment, but that can eventually be improved in a follow-up PR.. I have also added some tests, but we could as always test more. It's a huge PR already, but I still think it's reviewable. The tests might be a bit harder to review, but I tried to follow the established test-pattern - which is a bit complicated. 😅 

Please review and let me know what we are missing!

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -198,6 +198,20 @@ type CertificateRequestPolicyAllowedStringSlice struct {
...
 	Required *bool `json:"required,omitempty"`
+
+	// Validations applies rules using Common Expression Language (CEL) to
+	// validate attribute values present on request beyond what is possible
+	// to express using values/required.
+	// ALL attribute values on the related CertificateRequest field must pass
+	// ALL validations for the request to be granted by this policy.
+	//
+	// NOTE: Validations will only be performed after a request is considered
+	// granted by this policy values/required fields. This behavior might
```
jsoref, line 209:
```suggestion
	// granted by this policy's values/required fields. This behavior might
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -198,6 +198,20 @@ type CertificateRequestPolicyAllowedStringSlice struct {
...
+	// change in a future version of the API.
```
jsoref, line 210:
```suggestion
	// change in a future version of this API.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -219,6 +233,44 @@ type CertificateRequestPolicyAllowedString struct {
...
 	Required *bool `json:"required,omitempty"`
+
+	// Validations applies rules using Common Expression Language (CEL) to
+	// validate attribute value present on request beyond what is possible
+	// to express using value/required.
+	// An attribute value on the related CertificateRequest field must pass
+	// ALL validations for the request to be granted by this policy.
+	//
+	// NOTE: Validations will only be performed after a request is considered
+	// granted by this policy value/required fields. This behavior might
```
jsoref, line 244:
```suggestion
	// granted by this policy's value/required fields. This behavior might
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -219,6 +233,44 @@ type CertificateRequestPolicyAllowedString struct {
...
+	// change in a future version of the API.
```
jsoref, line 245:
```suggestion
	// change in a future version of this API.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -219,6 +233,44 @@ type CertificateRequestPolicyAllowedString struct {
...
+	// To enable more advanced validation rules, approver-policy provides the
+	// `cr` (map) variable to the CEL expression containing `namespace` and
+	// `name` of the `CertificateRequest` resource.
+	//
+	// Examples:
+	// - Rule for namespaced DNSNames:
+	//	 {"rule": "self.endsWith(cr.namespace + '.svc.cluster.local'"}
+	Rule string `json:"rule"`
+
+	// Message represents the message displayed when validation fails.
```
jsoref, line 267: Is this an output or an input?

`represents` feels like an output, but I think it's an input.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -219,6 +233,44 @@ type CertificateRequestPolicyAllowedString struct {
...
+	// The message is required if the Rule contains line breaks. The message
+	// must not contain line breaks.
```
jsoref, line 269:
```suggestion
	// Message is required if the Rule contains line breaks. Note that Message
	// must not contain line breaks.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -219,6 +233,44 @@ type CertificateRequestPolicyAllowedString struct {
...
+	// If unset, the message is "failed rule: {Rule}".
```
jsoref, line 270:
```suggestion
	// If unset, a fallback message is used: "failed rule: {Rule}".
```

pkg/internal/approver/allowed/validation.go
```diff
@@ -76,14 +76,28 @@ func (a allowed) Validate(_ context.Context, policy *policyapi.CertificateReques
 	}
 
 	for _, stringSlice := range stringSlices {
-		if stringSlice.slice != nil && stringSlice.slice.Required != nil && *stringSlice.slice.Required && stringSlice.slice.Values == nil {
-			el = append(el, field.Required(stringSlice.path.Child("values"), "values must be defined if required field"))
+		if stringSlice.slice != nil {
+			if stringSlice.slice.Required != nil && *stringSlice.slice.Required && stringSlice.slice.Values == nil {
+				el = append(el, field.Required(stringSlice.path.Child("values"), "values must be defined if required field"))
```
jsoref, line 81: i know you aren't changing this text, but I don't like this text :(

pkg/internal/approver/allowed/validation_test.go
```diff
@@ -44,7 +44,7 @@ func Test_Validate(t *testing.T) {
 				Errors:  nil,
 			},
 		},
-		"if policy contains validation errors, expect a Allowed=false response": {
+		"if policy contains 'required' validation errors, expect a Allowed=false response": {
```
jsoref, line 47:
```suggestion
		"if policy contains 'required' validation errors, expect an Allowed=false response": {
```

pkg/internal/approver/allowed/validation_test.go
```diff
@@ -139,11 +139,79 @@ func Test_Validate(t *testing.T) {
 				Errors:  nil,
 			},
 		},
+		"if policy contains invalid CEL validations, expect a Allowed=false response": {
```
jsoref, line 142:
```suggestion
		"if policy contains invalid CEL validations, expect an Allowed=false response": {
```

pkg/internal/approver/validation/cache.go
```diff
@@ -0,0 +1,45 @@
+package validation
+
+// Cache maintains a cache of compiled validators.
+// The current implementation is a simple lazy cache meaning:
+// Whenever a validator is requested, it first checks the cache. If a compiled
+// validator exists for the supplied CEL expression, it is returned.
+// If the validator doesn't exist in the cache, a new validator is created
+// and compiled. Then added to the cache and returned.
```
jsoref, line 8:
```suggestion
// The current implementation is a simple lazy cache meaning:
//
// 1. Whenever a validator is requested, it first checks the cache.
// 2. If a compiled validator exists for the supplied CEL expression, it is returned.
// 3. If the validator doesn't exist in the cache, a new validator is created, compiled, added to the cache, and returned.
```

I'm not wed to this notation, I just have trouble w/ `meaning:` leading into a strange fully formed paragraph.

pkg/internal/approver/validation/validator.go
```diff
@@ -0,0 +1,87 @@
...
+type Validator interface {
+	// Compile compiles the Validator CEL expression into a runnable program
+	// that can perform validations. Any compilation errors will be returned.
+	Compile() error
+
+	// Validate validates the supplied value against the Validator CEL
+	// expression in context of the request.
+	// Returns 'true' if the value is valid (passes validation).
+	// A returned error should be considered as an internal/technical error,
+	// and not typically user-facing.
```
jsoref, line 31: I presume the goal here is to say "don't show this error to the user, it isn't human-readable", but I don't think that this prose gets you there.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: In the policy API it is an input, but it sets a message that will provide output to the user in case the rule fails. 😉

pkg/internal/approver/allowed/validation.go
erikgb: I agree, but I am trying hard to not get an XXXXXXXXL PR. 😉

pkg/internal/approver/validation/validator.go
erikgb: You got it. I will try to rephrase this.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
jsoref:
```suggestion
	// Message is the message to display when validation fails.
```

inteon (commented):
Sorry, I forgot to submit these two pending comments

pkg/internal/approver/validation/validator.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (v *validator) Compile() error {
+	if v.program != nil {
+		// Already compiled
+		return nil
+	}
+
+	env, err := cel.NewEnv(
+		cel.Variable(varSelf, cel.StringType),
+		cel.Variable(varRequest, cel.MapType(cel.StringType, cel.StringType)),
```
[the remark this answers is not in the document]
inteon: ^^ I think you already said that we could do this in a future PR?

pkg/internal/approver/allowed/evaluator_test.go
```diff
@@ -331,11 +331,107 @@ func Test_Evaluate(t *testing.T) {
...
+				noErrModifier(func(csr *x509.CertificateRequest) { csr.Subject.OrganizationalUnit = []string{"org-1", "org-2"} }),
+				noErrModifier(func(csr *x509.CertificateRequest) { csr.Subject.Locality = []string{"loc-1", "loc-2"} }),
+				noErrModifier(func(csr *x509.CertificateRequest) { csr.Subject.Province = []string{"prov-1", "prov-2"} }),
+				noErrModifier(func(csr *x509.CertificateRequest) { csr.Subject.StreetAddress = []string{"street-1", "street-2"} }),
+				noErrModifier(func(csr *x509.CertificateRequest) { csr.Subject.PostalCode = []string{"post-1", "post-2"} }),
+				noErrModifier(func(csr *x509.CertificateRequest) { csr.Subject.SerialNumber = "serial-1" }),
+			)), gen.SetCertificateRequestNamespace("foo")),
+			policy: policyapi.CertificateRequestPolicySpec{
+				Allowed: &policyapi.CertificateRequestPolicyAllowed{
+					CommonName:     &policyapi.CertificateRequestPolicyAllowedString{Validations: []policyapi.ValidationRule{{Rule: "self.contains('cn-1')"}}, Value: pointer.String("*")},
```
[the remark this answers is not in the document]
erikgb: Yes, I agree it's annoying. But then we are probably returning to the impossible condition discussions again. The existing logic in the validating webhook is troublesome...

pkg/internal/approver/validation/validator.go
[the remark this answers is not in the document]
erikgb: Believe me, I tried to get the protobuf-thing to work, but no luck so far. No matter how I did it, the CEL compiler didn't manage to find my type. And the CEL documentation is not awsome. If you can help me, I would appreciate it. It would be optimal to have the `cr` variable of type `metav1.ObjectMeta` - and the protos are already there. I just cannot make them work in CEL! 😡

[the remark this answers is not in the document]
erikgb: I already spent significant time trying to get the protos to work, so if I cannot get any help, I suggest staying with Map for now and improve it in a follow-up PR. Others might have more experience with protos than me.

pkg/internal/approver/allowed/validation.go
erikgb: I would prefer to not change unrelated stuff in this already huge PR. Hope it's ok.

pkg/internal/approver/allowed/evaluator_test.go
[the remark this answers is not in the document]
erikgb: I have now changed the PR to allow a user top specify [any of](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#anyof) value/values and validations. This also changes the webhook validation: if `required: true`, then any of value/values and validations MUST be specified. Kept this in a separate commit for now: https://github.com/cert-manager/approver-policy/pull/277/commits/8a4378cb8e711cb4c162ae5164d4c6b0ef8f34d3. Please let me know what you think.

pkg/internal/approver/validation/validator.go
erikgb: I have now made an attempt to rewrite this. PTAL.

pkg/internal/approver/allowed/evaluator_test.go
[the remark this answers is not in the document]
inteon: I like the values or validations approach.
What happens now if you specify both values and validations?

inteon:
/approve
I approve of this PR, but it needs some more work (definitely from my side) before I can lgtm.

erikgb:
/hold

I want at least to rebase commits before we merge.

pkg/internal/approver/allowed/evaluator_test.go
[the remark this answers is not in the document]
erikgb: If you specify both, both will apply. I should probably add a test for it.

pkg/internal/approver/validation/validator.go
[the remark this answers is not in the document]
erikgb: I think I finally made it work - at least locally on my workstation. 😊 CI seems to struggle finding the correct release of `protoc` to download. Any pointers are appreciated @inteon. See https://github.com/cert-manager/approver-policy/pull/277/commits/046b3624e8cb1ee2fc8e9e4d8f672ef58827a821.

pkg/internal/approver/allowed/evaluator_test.go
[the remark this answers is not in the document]
erikgb: I have now added a test for value/values AND validations. And glad I did. Caught a noob bug. 😊

pkg/internal/approver/validation/validator.go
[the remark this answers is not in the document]
erikgb: By replacing `uname -p` with `uname -m` in the detect-OS Makefile, I was able to get a step further. Now the problem is no `unzip` in the container? that runs CI. This should be possible to fix, but how? I am far outside my comfort zone here. Please help! 🙏

pkg/internal/approver/validation/validator.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+// Validator knows how to validate CSR attribute values in CertificateRequests
+// against CEL expressions declared in CertificateRequestPolicy.
+// A Validator must be compiled before it can validate. A compiled Validator
+// is stateless, thread-safe, and cacheable.
+type Validator interface {
+	// Compile compiles the Validator CEL expression into a runnable program
+	// that can perform validations. Any compilation errors will be returned.
+	Compile() error
```
inteon, line 25: @erikgb I think we can still improve this interface. I see two approaches:
- Compile() could return a new struct that has a Validate function.
- Validate() could lazily call Compile() when necessary & we can make the Compile() function private

Currently you have to remember whether Compile was called already, making it harder to use this interface.

Please let me know what you think.

erikgb: For now, we could remove `Compile` from the interface, as the cache compiles the validator (if new validator) before returning it to the caller. Your first suggestion will force us to put names on TWO interfaces, and we all know that naming is the hardest part of programming. 😉 The webhook will typically only compile the expression, to verify it's valid, and never invoke `Validate`. So I don't think your second suggestion is a good one - mainly because you suggest hiding away `Compile`. I generally think we should separate `Compile` from `Validate`,  mainly because the error handling must be different for those two operations.

pkg/internal/approver/validation/validator.go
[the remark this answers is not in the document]
erikgb: This should be resolved now: using a protobuf message as variable in expressions.

pkg/internal/approver/validation/validator.go
erikgb: Removed `Compile` from `Validator` interface.

pkg/internal/approver/validation/cache.go
```diff
@@ -0,0 +1,53 @@
...
+type Cache interface {
+	// Get returns a compiled validator for the supplied CEL expression.
+	// Any compilation errors will be returned to the caller.
+	//
+	// The supplied CEL expression must output a bool.
+	Get(expr string) (Validator, error)
+}
+
+type cache struct {
+	m *sync.Map
```
inteon, line 20:
```suggestion
	m sync.Map
```
I don't think this has to be a pointer to a sync.Map.

pkg/internal/approver/validation/cache.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+type cacheEntry struct {
+	validator *validator
+	err       error
+}
+
+func (c cache) Get(expr string) (Validator, error) {
```
inteon, line 28:
```suggestion
func (c *cache) Get(expr string) (Validator, error) {
```
This should probably be a function on the cache pointer instead of the non-pointer.

pkg/internal/approver/validation/validator.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Validator knows how to validate CSR attribute values in CertificateRequests
+// against CEL expressions declared in CertificateRequestPolicy.
+// Validator is stateless, thread-safe, and cacheable.
+type Validator interface {
+	// Validate validates the supplied value against the Validator CEL
+	// expression in context of the request.
+	// Returns 'true' if the value is valid (passes validation).
+	// Returned errors should be considered as internal/technical errors,
+	// and should NOT be returned unprocessed to end-users of the API.
```
jsoref, line 26: "unprocessed" is doing a lot of work here, but, it should (technically) be enough.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -180,13 +180,12 @@ type CertificateRequestPolicyAllowedX509Subject struct {
 }
 
 // CertificateRequestPolicyAllowedStringSlice represents allowed string values
-// paired with whether the field is a required value on the request.
+// and/or validations paired with whether the field is a required value on the request.
+// If no allowed values nor validations are specified, the related field must be empty.
```
jsoref, line 184:
```suggestion
// If neither allowed values nor validations are specified, the related field must be empty.
```

pkg/internal/approver/allowed/evaluator.go
```diff
@@ -164,79 +168,106 @@ func (e evaluator) Subject() subjectEvaluator {
...
-			el = append(el, field.Invalid(fldPath.Child("value"), s, "nil"))
-		} else if !util.WildcardMatches(*crp.Value, s) {
-			el = append(el, field.Invalid(fldPath.Child("value"), s, *crp.Value))
+func (a allowed) evaluateString(request *cmapi.CertificateRequest, s string, crp *policyapi.CertificateRequestPolicyAllowedString, fldPath *field.Path) field.ErrorList {
+	if len(s) == 0 {
+		if crp != nil && crp.Required != nil && *crp.Required {
+			return []*field.Error{field.Required(fldPath.Child("required"), strconv.FormatBool(*crp.Required))}
+		} else {
+			return nil
 		}
```
inteon, line 225:
```suggestion
		}
		return nil
```

Small style change.

pkg/internal/approver/allowed/evaluator.go
```diff
@@ -164,79 +168,106 @@ func (e evaluator) Subject() subjectEvaluator {
...
 
-func evaluateString(s string, crp *policyapi.CertificateRequestPolicyAllowedString, fldPath *field.Path) field.ErrorList {
-	var el field.ErrorList
-	if len(s) > 0 {
-		if crp == nil || crp.Value == nil {
-			el = append(el, field.Invalid(fldPath.Child("value"), s, "nil"))
-		} else if !util.WildcardMatches(*crp.Value, s) {
-			el = append(el, field.Invalid(fldPath.Child("value"), s, *crp.Value))
+func (a allowed) evaluateString(request *cmapi.CertificateRequest, s string, crp *policyapi.CertificateRequestPolicyAllowedString, fldPath *field.Path) field.ErrorList {
+	if len(s) == 0 {
```
inteon, line 220: Could you add some comments in this function?
eg. explain why if len(s) == 0 we do not care about crp.Value or crp.Validations.

pkg/internal/approver/allowed/validation.go
```diff
@@ -76,14 +76,32 @@ func (a allowed) Validate(_ context.Context, policy *policyapi.CertificateReques
 	}
 
 	for _, stringSlice := range stringSlices {
-		if stringSlice.slice != nil && stringSlice.slice.Required != nil && *stringSlice.slice.Required && stringSlice.slice.Values == nil {
-			el = append(el, field.Required(stringSlice.path.Child("values"), "values must be defined if required field"))
+		if stringSlice.slice != nil {
+			if stringSlice.slice.Required != nil && *stringSlice.slice.Required {
+				if stringSlice.slice.Values == nil && len(stringSlice.slice.Validations) == 0 {
+					el = append(el, field.Required(stringSlice.path.Child("values"), "values/validations must be defined if required field"))
```
jsoref, line 82:
```suggestion
					el = append(el, field.Required(stringSlice.path.Child("values"), "values/validations must be defined if field is required"))
```
? or is this the "don't show this error to the user, it isn't human-readable" bit?

inteon (commented):
Thanks @erikgb this code looks good to go.
I just added some final comments wrt. some code style changes & need for some more comments.
Great work.

pkg/internal/approver/validation/cache.go
```diff
@@ -0,0 +1,53 @@
...
+	o, ok := c.m.Load(expr)
+	if ok {
+		ce := o.(*cacheEntry)
+		return ce.validator, ce.err
+	}
+
+	// Expression did not exist in cache. Create a new validator, compile it
+	// and add the result to cache.
+	// Theoretically this could lead to the same expression being compiled multiple times,
+	// but guarding against that would require lock increasing complexity.
```
jsoref, line 39:
```suggestion
	// but guarding against that would require locking and increase complexity.
```

pkg/internal/approver/validation/cache.go
erikgb: Thanks! Fixed

pkg/internal/approver/allowed/validation.go
inteon, line 82:
```suggestion
					el = append(el, field.Required(stringSlice.path.Child("values"), "'values' or 'validations' must be defined if required field"))
```

erikgb: No, this is a user-facing error. 😉 I think the suggestion from @inteon is even better.

pkg/internal/approver/allowed/evaluator.go
```diff
@@ -247,3 +277,24 @@ func evaluateBool(b bool, crp *bool, fldPath *field.Path) field.ErrorList {
...
+			el = append(el, field.InternalError(fldPath.Index(i), err))
+			continue
+		}
+		valid, err := validator.Validate(s, *request)
+		if err != nil {
+			el = append(el, field.InternalError(fldPath.Index(i), err))
+			continue
+		}
+		if !valid {
+			detail := ptr.Deref(v.Message, fmt.Sprintf("failed rule: %s", v.Rule))
```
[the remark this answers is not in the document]
inteon: Also, do we have to quote the rule to make the output more readable?

pkg/internal/approver/validation/validator.go
erikgb: I know, but this feature could be further improved. This PR should be considered an MVP.

pkg/internal/approver/allowed/validation.go
```diff
@@ -76,14 +76,32 @@ func (a allowed) Validate(_ context.Context, policy *policyapi.CertificateReques
 	}
 
 	for _, stringSlice := range stringSlices {
-		if stringSlice.slice != nil && stringSlice.slice.Required != nil && *stringSlice.slice.Required && stringSlice.slice.Values == nil {
-			el = append(el, field.Required(stringSlice.path.Child("values"), "values must be defined if required field"))
+		if stringSlice.slice != nil {
+			if stringSlice.slice.Required != nil && *stringSlice.slice.Required {
+				if stringSlice.slice.Values == nil && len(stringSlice.slice.Validations) == 0 {
+					el = append(el, field.Required(stringSlice.path.Child("values"), "'values' or 'validations' must be defined if field is 'required'"))
```
erikgb, line 82: A user can specify `values` **AND** `validations`. Does this error message express this as it stands now, or do we have an even better suggestion? 🤔

inteon: maybe or/and could work, but I think the current message is also correct (the issue in this case is that at least one of the two must be specified)

pkg/internal/approver/allowed/evaluator.go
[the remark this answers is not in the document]
erikgb: I can add a test, but regarding quoting I tried to follow upstream as close as possible:

https://github.com/kubernetes/kubernetes/blob/56d7898510f2a973f92fda13c2ba3a5e756d9621/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel/validation.go#L629

pkg/internal/approver/allowed/validation.go
erikgb: @jsoref Can you help here? :-)

pkg/internal/approver/allowed/evaluator.go
erikgb: Done. Might not be the best wording, but this has more to do with not breaking the policy API.

pkg/internal/approver/allowed/validation.go
```diff
@@ -76,14 +76,32 @@ func (a allowed) Validate(_ context.Context, policy *policyapi.CertificateReques
...
 		}
 	}
 
 	for _, stringI := range strings {
-		if stringI.string != nil && stringI.string.Required != nil && *stringI.string.Required && stringI.string.Value == nil {
-			el = append(el, field.Required(stringI.path.Child("value"), "value must be defined if required field"))
+		if stringI.string != nil {
+			if stringI.string.Required != nil && *stringI.string.Required {
+				if stringI.string.Value == nil && len(stringI.string.Validations) == 0 {
+					el = append(el, field.Required(stringI.path.Child("value"), "value/validations must be defined if required field"))
```
inteon, line 97: same as above

pkg/internal/approver/allowed/evaluator.go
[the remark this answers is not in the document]
erikgb: We already have many tests for this "failed rule: " case. Please take another look. 😉

pkg/internal/approver/allowed/validation.go
erikgb: Done.

inteon (commented):
Thanks a lot for this great improvement!
/approve
/lgtm
/hold in case you want to add more info

erikgb:
Thanks @inteon! I will allow @jsoref to do a final pass and update the design PR (https://github.com/cert-manager/approver-policy/pull/256) before we merge. Looking forward to this!

pkg/internal/approver/allowed/validation.go
jsoref: "at least one of ..."?

erikgb: Updated.

erikgb:
I don't think I have more to add to this at this point.

/unhold

PR #280, opened 2023-10-06
docs: refine CertificateRequestPolicy API docs

erikgb:
I find the API docs generated from Go docs hard to read to understand. IMO the API docs should be for users of the Kubernetes API (CRD) - that could be using any programming language, and not written like traditional Go API docs.

This PR is an attempt to improve the docs with a focus on the generated field descriptions in the CRD OpenAPI specification. It might look like a lot has changed, but remember that changes in the API types propagate to generated files.

A high-level overview of the proposed changes:

- Avoid the term ``nil` in CRD description fields and instead use well-known API terms as omitted and empty.
- Avoid repetition (on involved fields) of cross-field validations. E.g. `Values` and `Required`.
- Field validations and field behavior documented on the field, and not in the "outer" scope.

As English is not my native language, I might need some help in wording, but I hope we can focus on structure in the first rounds of review.


docs/api/api.md
```diff
@@ -135,42 +135,36 @@ type CertificateRequestPolicyAllowed struct {
...
     // +optional
     URIs *CertificateRequestPolicyAllowedStringSlice `json:"uris,omitempty"`
 
-    // EmailAddresses defines the X.509 Email SANs that may be
-    // requested for.
+    // EmailAddresses defines the X.509 Email SANs that may be requested for.
     // +optional
     EmailAddresses *CertificateRequestPolicyAllowedStringSlice `json:"emailAddresses,omitempty"`
 
     // IsCA defines whether it is permissible for a CertificateRequest to have
```
inteon, line 153: I don't really understand why we are using the word `permissible` so often in these docs. Shouldn't we just use "whether it is permitted" or "whether it is allowed"?

docs/api/api.md
```diff
@@ -135,42 +135,36 @@ type CertificateRequestPolicyAllowed struct {
...
+    // Only a value of `true` permits CertificateRequests setting the `spec.isCA`
+    // field to `true`.
     // +optional
     IsCA *bool `json:"isCA,omitempty"`
 
     // Usages defines the list of permissible key usages that may appear
     // on the CertificateRequest `spec.keyUsages` field.
-    // An omitted field or value of `nil` forbids any Usages being requested.
-    // An empty slice `[]` is equivalent to `nil`.
+    // An omitted field or empty list forbids any key usages being requested.
```
inteon, line 162: Could we rephrase this sentence so it is even easier to understand (eg. https://github.com/cert-manager/cert-manager/blob/master/pkg/apis/certmanager/v1/types_certificate.go#L252-L254)?

docs/api/api.md
jsoref:
```suggestion
IsCA defines if a CertificateRequest is allowed to have
```

erikgb: Thanks @jsoref! I have now "washed" the API docs for the "permissible" term. I hope it's an improvement. 😉

docs/api/api.md
erikgb: @inteon I am unsure if I can improve the text with the link provided. In this case, it's a pointer to a slice of `KeyUsage` (alias for `string`), which is a bit different. I would prefer to avoid the term `nil`, which is a Go concept - and does not make sense IMO to use in Kubernetes API docs for yaml/json documents.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,92 +53,90 @@ type CertificateRequestPolicyList struct {
...
-	// IsCA defines whether it is permissible for a CertificateRequest to have
-	// the `spec.IsCA` field set to `true`.
-	// An omitted field, value of `nil` or `false`, forbids the `spec.IsCA` field
-	// from bring `true`.
-	// A value of `true` permits CertificateRequests setting the `spec.IsCA` field
-	// to `true`.
+	// IsCA defines if a CertificateRequest is allowed to have the `spec.isCA`
+	// field set to `true`.
+	// Only a value of `true` permits CertificateRequests setting the `spec.isCA`
+	// field to `true`.
```
inteon, line 116:
```suggestion
	// If `true`, the `spec.isCA` field can be `true` or `false`
	// If `false` or unset, the `spec.isCA` field must be `false`
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -178,102 +176,99 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 }
 
-// CertificateRequestPolicyAllowedStringSlice represents an allowed string
-// slice value paired with whether the field is a required value on the
-// request.
+// CertificateRequestPolicyAllowedStringSlice represents allowed string values paired
+// with whether the field is a required value on the request.
+// An empty Values paired with Required `true` is an impossible condition that
+// always denies.
+// Values may not be omitted` if Required is `true`.
```
inteon, line 183:
```suggestion
// The validating webhook enforces that Values can not be omitted if Required is `true`.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -178,102 +176,99 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
+	// request from being requested.
 	// TODO: add x-kubernetes-list-type: set in v1alpha2
 	// +optional
 	Values *[]string `json:"values,omitempty"`
 
-	// Required marks this field as being a required value on the request.
-	// May only be set to true if Values is also defined.
-	// Default is nil which marks the field as not required.
+	// Required defines if this field is a required value on the request.
+	// Set to `true` to mark the field as required on request.
```
inteon, line 195:
```suggestion
	// Required defines if the related field must have at least one value.
	// Defaults to `false`.
```

docs/api/api.md
erikgb: I think this should be ok now. Please add suggestions to the source file: types_certificaterequestpolicy.go. 😉

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -178,102 +177,101 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
-	// Value defines the value that is permissible to be present on the request.
+	// Value defines the allowed attribute value on the related
+	// CertificateRequest field.
 	// Accepts wildcards "*".
-	// An omitted field or value of `nil` forbids the value from being requested.
-	// An empty string is equivalent to `nil`, however an empty string paired with
-	// Required as `true` is an impossible condition that always denies.
-	// Value may not be `nil` if Required is `true`.
+	// An omitted or empty string forbids the value on the related field in the
+	// request from being requested.
```
inteon, line 213:
```suggestion
	// If set, the related field must match the specified pattern.
	// If `""` or unset, the related field must be empty.
	//
	// NOTE: If `Required` is `true`, this field cannot be omitted. Setting the field to `""` instead, results in an always-false expression.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -178,102 +177,101 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 	// +optional
 	Value *string `json:"value,omitempty"`
 
 	// Required marks this field as being a required value on the request.
```
inteon, line 217:
```suggestion
	// Required marks that the related field must be provided and not be an empty string.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,95 +53,96 @@ type CertificateRequestPolicyList struct {
 // CertificateRequestPolicySpec defines the desired state of
 // CertificateRequestPolicy.
 type CertificateRequestPolicySpec struct {
-	// Allowed is the set of attributes that are "allowed" by this policy. A
-	// CertificateRequest will only be considered permissible for this policy if
-	// the CertificateRequest has the same or less as what is allowed.  Empty or
-	// `nil` allowed fields mean CertificateRequests are not allowed to have that
-	// field present to be permissible.
+	// Allowed defines the attributes a CertificateRequest is allowed to have.
```
jsoref, line 56:
```suggestion
	// Allowed defines the allowed attributes for a CertificateRequest.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,95 +53,96 @@ type CertificateRequestPolicyList struct {
...
+	// field _must_ be omitted or have an empty value for the request to be
+	// permitted.
 	// +optional
 	Allowed *CertificateRequestPolicyAllowed `json:"allowed,omitempty"`
 
-	// Constraints is the set of attributes that _must_ be satisfied by the
-	// CertificateRequest for the request to be permissible by the policy. Empty
-	// or `nil` constraint fields mean CertificateRequests satisfy that field
-	// with any value of their corresponding attribute.
+	// Constraints define fields that _must_ be satisfied by the
```
jsoref, line 66:
```suggestion
	// Constraints define fields that _must_ be satisfied by a
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,95 +53,96 @@ type CertificateRequestPolicyList struct {
...
+	// CertificateRequest for the request to be allowed by this policy.
+	// Omitted fields will be satisfied by any value on the corresponding
+	// attribute on the request.
```
jsoref, line 69:
```suggestion
	// Omitted fields place no restrictions on the corresponding
	// attribute in a request.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,95 +53,96 @@ type CertificateRequestPolicyList struct {
...
 	// +optional
 	Constraints *CertificateRequestPolicyConstraints `json:"constraints,omitempty"`
 
-	// Plugins define a set of plugins and their configuration that should be
-	// executed when this policy is evaluated against a CertificateRequest. A
-	// plugin must already be built within approver-policy for it to be
+	// Plugins define plugins and their configuration that should be
+	// executed when this policy is evaluated against a CertificateRequest.
+	// A plugin must already be built within approver-policy for it to be
 	// available.
```
jsoref, line 76: This isn't enough information for me to do something.

Is a plugin some code I had to compile? Is it something I had to define elsewhere? Where? Is it another CRD?

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,95 +53,96 @@ type CertificateRequestPolicyList struct {
...
 
-// CertificateRequestPolicyAllowed is a set of attributes that are declared as
-// permissible for a CertificateRequest to have those values present. It is
-// permissible for a CertificateRequest to request _less_ than what is allowed,
-// but _not more_, i.e. it is permissible for a CertificateRequest to request a
-// subset of what is allowed.
-// Empty fields or `nil` values declares that the equivalent CertificateRequest
-// field _must_ be omitted or empty for the request to be permitted.
+// CertificateRequestPolicyAllowed defines the attributes a CertificateRequest
+// is allowed to have.
```
jsoref, line 87: Suggested change
```suggestion
// CertificateRequestPolicyAllowed defines the allowed attributes for a
// CertificateRequest.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,95 +53,96 @@ type CertificateRequestPolicyList struct {
...
+// A CertificateRequest can request _less_ than what is allowed,
+// but _not more_, i.e. a CertificateRequest can request a subset of what is
+// declared as allowed  by the policy.
+// Omitted fields declares that the equivalent CertificateRequest field _must_
+// be omitted or have an empty value for the request to be permitted.
```
jsoref, line 92: I kinda feel like this should have an `Unlike CertificateRequestPolicySpec, ...`

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,95 +53,96 @@ type CertificateRequestPolicyList struct {
...
 type CertificateRequestPolicyAllowed struct {
-	// CommonName defines the X.509 Common Name that is permissible.
+	// CommonName defines the X.509 Common Name a CertificateRequest is
+	// allowed to have.
```
jsoref, line 95: This feels wrong, presumably it's the format or a pattern? It can't just be a single value. (But otherwise, the style I've outlined above...)

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,95 +53,96 @@ type CertificateRequestPolicyList struct {
...
 	// +optional
 	CommonName *CertificateRequestPolicyAllowedString `json:"commonName,omitempty"`
 
 	// DNSNames defines the X.509 DNS SANs that may be requested for.
```
jsoref, line 99: Some text explaining how this relates to the one above would be nice...

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,95 +53,96 @@ type CertificateRequestPolicyList struct {
...
-	// Accepts wildcards "*".
 	// +optional
 	DNSNames *CertificateRequestPolicyAllowedStringSlice `json:"dnsNames,omitempty"`
 
-	// IPAddresses defines the X.509 IP SANs that may be requested
-	// for.
+	// IPAddresses defines the X.509 IP SANs that may be requested for.
```
jsoref, line 103:
```suggestion
	// IPAddresses defines the X.509 IP SANs that may be requested.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,95 +53,96 @@ type CertificateRequestPolicyList struct {
...
 	// +optional
 	IPAddresses *CertificateRequestPolicyAllowedStringSlice `json:"ipAddresses,omitempty"`
 
 	// URIs defines the X.509 URI SANs that may be requested for.
 	// +optional
 	URIs *CertificateRequestPolicyAllowedStringSlice `json:"uris,omitempty"`
 
-	// EmailAddresses defines the X.509 Email SANs that may be
-	// requested for.
+	// EmailAddresses defines the X.509 Email SANs that may be requested for.
```
jsoref, line 111:
```suggestion
	// EmailAddresses defines the X.509 Email SANs that may be requested.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,95 +53,96 @@ type CertificateRequestPolicyList struct {
...
 	EmailAddresses *CertificateRequestPolicyAllowedStringSlice `json:"emailAddresses,omitempty"`
 
-	// IsCA defines whether it is permissible for a CertificateRequest to have
-	// the `spec.IsCA` field set to `true`.
-	// An omitted field, value of `nil` or `false`, forbids the `spec.IsCA` field
-	// from bring `true`.
-	// A value of `true` permits CertificateRequests setting the `spec.IsCA` field
-	// to `true`.
+	// IsCA defines if a CertificateRequest is allowed to have the `spec.isCA`
+	// field set to `true`.
```
jsoref, line 116:
```suggestion
	// IsCA defines whether or not a CertificateRequest is allowed to set the `spec.isCA`
	// field set to `true`.
```

jsoref (commented):
I'm checkpointing, more in a bit

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: Yes, a plugin is built into approver-policy at build-time. I have tried to explain this in my experimental plugin adding CEL validation support: https://github.com/erikgb/cel-approver-policy-plugin#installation

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
 	// +optional
 	PostalCodes *CertificateRequestPolicyAllowedStringSlice `json:"postalCodes,omitempty"`
 
-	// SerialNumber defines the X.509 Subject Serial Number that may be requested
-	// for.
+	// SerialNumber defines the X.509 Subject Serial Number that may be
+	// requested for.
```
jsoref, line 178:
```suggestion
	// SerialNumber defines the X.509 Subject Serial Number that may be
	// requested.
```

This is weird. What kind of format is this? If it's a single value, that makes no sense...

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 	SerialNumber *CertificateRequestPolicyAllowedString `json:"serialNumber,omitempty"`
 }
 
-// CertificateRequestPolicyAllowedStringSlice represents an allowed string
-// slice value paired with whether the field is a required value on the
-// request.
+// CertificateRequestPolicyAllowedStringSlice represents allowed string values
+// paired with whether the field is a required value on the request.
+// An empty Values paired with Required `true` is an "always-false" condition.
+// always denies.
```
jsoref, line 186: This line is misplaced or missing words or the punctuation on the line before it is wrong

jsoref (commented):
checkpointing

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: A policy can literally accept just a single common name value. But the field supports wildcards. Not sure about the style you are referring to, sorry. Can you please elaborate?

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
-	// Defines the values that are permissible to be present on request.
+	// Values defines allowed attribute values on the related
+	// CertificateRequest field.
 	// Accepts wildcards "*".
-	// An omitted field or value of `nil` forbids any value on the related field
-	// in the request from being requested.
-	// An empty slice `[]` is equivalent to `nil`, however an empty slice paired
-	// with Required `true` is an impossible condition that always denies.
-	// Values may not be `nil` if Required is `true`.
+	// If set, the related field must contain a subset of the allowed values.
```
jsoref, line 193: Must it be a proper subset? Or does this just mean it can't be a superset / have items not in this list?

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
+	// If `[]` or unset, the related field must be empty.
+	//
+	// NOTE: If `Required` is `true`, this field cannot be omitted. Setting the
+	// field to `[]` instead, results in an always-false expression.
```
jsoref, line 197: Not sure what `always-false` means here -- does it mean this policy will always reject?

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
+	// Required defines if the related field must have at least one value.
+	// Defaults to `false`.
 	// +optional
 	Required *bool `json:"required,omitempty"`
 }
 
 // CertificateRequestPolicyAllowedString represents an allowed string value
 // paired with whether the field is a required value on the request.
+// An empty Value paired with Required `true` is an impossible condition that
+// always denies.
```
jsoref, line 211: This sounds like better text than the previous `always-false` text.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
+// Value may not be omitted if Required is `true`.
 type CertificateRequestPolicyAllowedString struct {
-	// Value defines the value that is permissible to be present on the request.
+	// Value defines the allowed attribute value on the related
+	// CertificateRequest field.
```
jsoref, line 215: the word `pattern` occurs below and probably answers some of my questions above -- it should be used more often.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 }
 
-// CertificateRequestPolicyConstraints define fields that, if defined, _must_
-// be satisfied by the CertificateRequest for the request to be permissible by
-// this policy. Fields that are omitted or have a value of `nil` will be
-// satisfied by any value on the corresponding attribute on the request.
+// CertificateRequestPolicyConstraints define fields that _must_ be satisfied
+// by the CertificateRequest for the request to be allowed by this policy.
+// Omitted fields will be satisfied by any value on the corresponding attribute
+// on the request.
```
jsoref, line 235:
```suggestion
// Omitted fields will be satisfied by any value in the corresponding attribute
// of the request.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 type CertificateRequestPolicyConstraints struct {
 	// MinDuration defines the minimum duration a certificate may be requested
 	// for.
```
jsoref, line 238:
```suggestion
	// MinDuration defines the minimum duration for a certificate request.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
-	// Values are inclusive (i.e. a min value of `1h` will accept a duration of
+	// Values are inclusive (i.e. a value of `1h` will accept a duration of
 	// `1h`). MinDuration and MaxDuration may be the same value.
-	// An omitted field or value of `nil` permits any minimum duration.
-	// If MinDuration is defined, a duration _must_ be requested on the
-	// CertificateRequest.
+	// If set, a duration _must_ be requested on the CertificateRequest.
+	// An omitted field implies no minimum constraint on duration.
```
jsoref, line 242:
```suggestion
	// An omitted field implies no minimum constraint for duration.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 	// +optional
 	MinDuration *metav1.Duration `json:"minDuration,omitempty"`
 
 	// MaxDuration defines the maximum duration a certificate may be requested
```
jsoref, line 246:
```suggestion
	// MaxDuration defines the maximum duration for a certificate request.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 type CertificateRequestPolicyConstraints struct {
 	// MinDuration defines the minimum duration a certificate may be requested
 	// for.
-	// Values are inclusive (i.e. a min value of `1h` will accept a duration of
+	// Values are inclusive (i.e. a value of `1h` will accept a duration of
 	// `1h`). MinDuration and MaxDuration may be the same value.
-	// An omitted field or value of `nil` permits any minimum duration.
-	// If MinDuration is defined, a duration _must_ be requested on the
-	// CertificateRequest.
+	// If set, a duration _must_ be requested on the CertificateRequest.
```
jsoref, line 241:
```suggestion
	// If set, a duration _must_ be requested in the CertificateRequest.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 	// MaxDuration defines the maximum duration a certificate may be requested
 	// for.
-	// Values are inclusive (i.e. a max value of `1h` will accept a duration of
-	// `1h`). MaxDuration and MinDuration may be the same value.
-	// An omitted field or value of `nil` permits any maximum duration.
-	// If MaxDuration is defined, a duration _must_ be requested on the
-	// CertificateRequest.
+	// Values are inclusive (i.e. a value of `1h` will accept a duration of
+	// `1h`). MinDuration and MaxDuration may be the same value.
+	// If set, a duration _must_ be requested on the CertificateRequest.
```
jsoref, line 250:
```suggestion
	// If set, a duration _must_ be requested in the CertificateRequest.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
+	// An omitted field implies no maximum constraint on duration.
```
jsoref, line 251: I'm not a fan of `implies`. I'd tentatively lean to `applies`:
```suggestion
	// An omitted field applies no maximum constraint to duration.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 	// +optional
 	MaxDuration *metav1.Duration `json:"maxDuration,omitempty"`
 
-	// PrivateKey defines the shape of permissible private keys that may be used
-	// for the request with this policy.
-	// An omitted field or value of `nil` permits the use of any private key by
-	// the requestor.
+	// PrivateKey defines constraints on what shape of private key a
+	// CertificateRequest is allowed to use for its request.
```
jsoref, line 256:
```suggestion
	// PrivateKey defines constraints on the shape of private key
	// allowed for a CertificateRequest.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
+	// An omitted field implies no private key shape constraints.
```
jsoref, line 257: Again, not a fan of `implies`

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 	// +optional
 	PrivateKey *CertificateRequestPolicyConstraintsPrivateKey `json:"privateKey,omitempty"`
 }
 
 // CertificateRequestPolicyConstraintsPrivateKey defines constraints on what
```
jsoref, line 262: See previous set of changes. I'm not a fan of `what`

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
-// shape of private key is permissible for a CertificateRequest to have used
-// for its request.
+// shape of private key a CertificateRequest is allowed to use for its request.
 type CertificateRequestPolicyConstraintsPrivateKey struct {
 	// Algorithm defines the allowed crypto algorithm that is used by the
 	// requestor for their private key in their request.
```
jsoref, line 266:
```suggestion
	// Algorithm defines the allowed crypto algorithm for the private key
	// in a request.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
-	// An omitted field or value of `nil` permits any Algorithm.
+	// An omitted field permits any Algorithm.
```
jsoref, line 267: should Algorithm be capitalized above?

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 	// +optional
 	Algorithm *cmapi.PrivateKeyAlgorithm `json:"algorithm,omitempty"`
 
-	// MinSize defines the minimum key size a requestor may use for their private
-	// key.
+	// MinSize defines the minimum key size a requestor may use for their
+	// private key.
```
jsoref, line 272:
```suggestion
	// MinSize defines the minimum key size for a private key.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
+	// NOTE: If `Required` is `true`, this field cannot be omitted. Setting the
+	// field to `[]` instead, results in an always-false expression.
 	// TODO: add x-kubernetes-list-type: set in v1alpha2
 	// +optional
 	Values *[]string `json:"values,omitempty"`
 
-	// Required marks this field as being a required value on the request.
-	// May only be set to true if Values is also defined.
-	// Default is nil which marks the field as not required.
+	// Required defines if the related field must have at least one value.
```
jsoref, line 202:
```suggestion
	// Required controls whether the related field must have at least one value.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 	Algorithm *cmapi.PrivateKeyAlgorithm `json:"algorithm,omitempty"`
 
-	// MinSize defines the minimum key size a requestor may use for their private
-	// key.
+	// MinSize defines the minimum key size a requestor may use for their
+	// private key.
 	// Values are inclusive (i.e. a min value of `2048` will accept a size
 	// of `2048`). MinSize and MaxSize may be the same value.
-	// An omitted field or value of `nil` permits any minimum size.
+	// An omitted field permits any minimum size.
```
jsoref, line 275: I can't keep track of whether multiple policies can apply concurrently w/ a requirement for all to pass... Assuming it's an all must pass, this wording change makes sense:

```suggestion
	// An omitted field implies no additional minimum constraint on size.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 	MinSize *int `json:"minSize,omitempty"`
 
-	// MaxSize defines the maximum key size a requestor may use for their private
-	// key.
+	// MaxSize defines the maximum key size a requestor may use for their
+	// private key.
 	// Values are inclusive (i.e. a min value of `2048` will accept a size
 	// of `2048`). MaxSize and MinSize may be the same value.
-	// An omitted field or value of `nil` permits any maximum size.
+	// An omitted field permits any maximum size.
```
jsoref, line 283:
```suggestion
	// An omitted field implies no additional maximum constraint on size.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -172,118 +174,123 @@ type CertificateRequestPolicyAllowedX509Subject struct {
...
 	// of `2048`). MinSize and MaxSize may be the same value.
-	// An omitted field or value of `nil` permits any minimum size.
+	// An omitted field permits any minimum size.
 	// +optional
 	MinSize *int `json:"minSize,omitempty"`
 
-	// MaxSize defines the maximum key size a requestor may use for their private
-	// key.
+	// MaxSize defines the maximum key size a requestor may use for their
+	// private key.
```
jsoref, line 280:
```suggestion
	// MaxSize defines the maximum key size for a private key.
```

jsoref (commented):
checkpointing

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -294,71 +301,67 @@ type CertificateRequestPolicyPluginData struct {
...
 
 // CertificateRequestPolicySelectorIssuerRef defines the selector for matching
-// on `issuerRef` of requests.
+// on issuer references of requests.
 type CertificateRequestPolicySelectorIssuerRef struct {
-	// Name is the wildcard selector to match the `spec.issuerRef.name` field on
-	// requests.
+	// Name is the wildcard selector to match the `spec.issuerRef.name` field
+	// on requests.
 	// Accepts wildcards "*".
```
jsoref, line 332: It feels like I need a thing that says `unlike the X which doesn't accept wildcards`.

jsoref: Is this correct?
```suggestion
	// Name is a wildcard enabled selector that matches the
	// `spec.issuerRef.name` field of requests.
	// Accepts wildcards "*".
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -294,71 +301,67 @@ type CertificateRequestPolicyPluginData struct {
...
+	// An omitted field matches all groups.
 	// +optional
 	Group *string `json:"group,omitempty"`
 }
 
 // CertificateRequestPolicySelectorNamespace defines the selector for matching
-// on the `Namespace` of requests. Note that all selectors in the Namespace
-// selector must match in order for the request to be considered for evaluation
-// by this policy.
+// on the namespace of requests. Note that all selectors must match in order
```
jsoref, line 353:
```suggestion
// the namespace of requests. Note that all selectors must match in order
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -294,71 +301,67 @@ type CertificateRequestPolicyPluginData struct {
...
+	// MatchNames is the set of namespace names that select on
+	// CertificateRequests that have been created in a matching namespace.
 	// Accepts wildcards "*".
 	// TODO: add x-kubernetes-list-type: set in v1alpha2
 	// +optional
 	MatchNames []string `json:"matchNames,omitempty"`
 
 	// MatchLabels is the set of Namespace labels that select on
-	// CertificateRequests which have been created in a Namespace matching the
+	// CertificateRequests which have been created in a namespace matching the
```
jsoref, line 364: This changes makes me even more curious about `Algorithm` above...

jsoref (commented):
I've hit the end of the file. Note that some of my comments likely apply beyond where I've left them.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: It is equivalent to the `CommonName` field. While the `CommonName` field might be used, I also think this field makes little sense. But we cannot remove it without breaking the API.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: > A subset is a set that has either some or all of the elements of another set, called the superset.

I am not sure what is the correct wording here. 🤔 Maybe "can only have values matching at least one item in the list"?

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: Yes, the wording here was originally suggested by @inteon. In the current docs, the term "impossible condition" is used. I think this is a well-known term in the policy engine domain. Maybe we should stick to that? WDYT @jsoref? Is it better or worse?

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb:
```suggestion
	// An omitted field permits any algorithm.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: > Again, not a fan of implies

😉

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: Ref. https://cert-manager.io/docs/policy/approval/approver-policy/#configuration, I don't think the suggested rewording is correct:
image

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: I have tried to reword this now.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: Reworded.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: The above is now reworded. Is anything more needed?

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: Slightly reworded. Not sure if I can improve more.

SgtCoDFish (commented):
Added a couple of comments. I think this is a huge improvement - just a couple of bits maybe before we merge.

Thanks to everyone involved in this - it's a marathon but there are some awesome improvements here!

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
jsoref: ok, so it isn't an _all_ (**and**), it's an _any_ (**or**) matching policy.

I think the text is still wrong, although I don't know how to fix it.

This is the `CertificateRequestPolicy`, right? And it's defining parameters for a `CertificateRequest`, right?
But the `CertificateRequest` will not have a `minimum size` field, it'll have a `size` field, which will just not have any particular minimum size, right?

```suggestion
	// An omitted field allows size to be as small as possible.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
jsoref: The problem is that it isn't permitting any maximum size, it's permitting the size to be as large as the requester likes.

```suggestion
	// An omitted field allows size to be as large as possible.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
jsoref: good enough

jsoref: Probably should be `applies` for consistency w/ others

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: Updated.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -178,102 +176,99 @@ type CertificateRequestPolicyAllowedX509Subject struct {
 	SerialNumber *CertificateRequestPolicyAllowedString `json:"serialNumber,omitempty"`
 }
 
-// CertificateRequestPolicyAllowedStringSlice represents an allowed string
-// slice value paired with whether the field is a required value on the
-// request.
+// CertificateRequestPolicyAllowedStringSlice represents allowed string values paired
+// with whether the field is a required value on the request.
+// An empty Values paired with Required `true` is an impossible condition that
```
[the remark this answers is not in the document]
jsoref: Hmm... maybe we should talk about `impossible`.

Is the idea that "you can't satisfy this requirement and thus this Policy will not apply, and thus unless some other policy does, your request will not be fulfilled"?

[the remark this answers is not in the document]
jsoref: If we could use "`Values`" instead of "Values", it'd be really helpful for readers, as it's really confusing -- as a reader, I expect to see `An empty value paired...` but what the text wants to do is talk about the `Values` field which is a different creature entirely.

```suggestion
// `Values: []` paired with `Required: true` establishes a policy that will never grant a `CertificateRequest`, but other policies may.
```

The reason I am proposing `will not grant` is that `deny` implies a final outcome, but that isn't what happens in cert-manager -- a `CertificateRequest` can be processed by any other `CertificateRequestPolicy`s and a `Certificate` issued by any that do _allow_ the `CertificateRequest`.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: @jsoref Could this remaining comment be misplaced somehow? Sorry if I'm slow, but if the location is correct I need some further explanations.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: Removed it now.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -294,71 +301,67 @@ type CertificateRequestPolicyPluginData struct {
...
-	// CertificateRequests. This policy will only be evaluated against a
-	// CertificateRequest whose `spec.issuerRef` field matches
-	// `spec.selector.issuerRef`. CertificateRequests will not be processed on
-	// unmatched `issuerRef` if defined, regardless of whether the requestor is
-	// bound by RBAC.
-	// Accepts wildcards "*".
-	// Omitted values are equivalent to "*".
+	// IssuerRef is used to select on CertificateRequest issuer reference,
+	// meaning the CertificateRequestPolicy will only match on
+	// CertificateRequests that refer to matching issuers.
```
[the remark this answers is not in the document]
erikgb: Reworded slightly.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
erikgb: No "implies" left now - except in the file heading.

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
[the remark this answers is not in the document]
erikgb: I am torn here. The suggestion from Josh is a bit long, and I am also in doubt about the suggested field value notation. WDYT @SgtCoDFish @inteon? Impossible condition, "always-false" or the longer alternative? Somehow I don't understand why any user would create a policy that won't grant any request. This is hard, and I need to stop working on this PR soon. 😅

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,312 +53,303 @@ type CertificateRequestPolicyList struct {
...
-// permissible for a CertificateRequest to request _less_ than what is allowed,
-// but _not more_, i.e. it is permissible for a CertificateRequest to request a
-// subset of what is allowed.
-// Empty fields or `nil` values declares that the equivalent CertificateRequest
-// field _must_ be omitted or empty for the request to be permitted.
+// CertificateRequestPolicyAllowed defines the allowed attributes for a
+// CertificateRequest.
+// A CertificateRequest can request _less_ than what is allowed,
+// but _not more_, i.e. a CertificateRequest can request a subset of what is
+// declared as allowed  by the policy.
```
jsoref, line 91:
```suggestion
// declared as allowed by the policy.
```

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
jsoref: Hmm, I can't recall. Just disregard it

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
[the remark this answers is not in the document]
erikgb: Ok, I have now made some (hopefully) "final" adjustments - based on the suggestion from @jsoref. I am hoping we are approaching the finish line of this marathon now. PTAL!

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -53,312 +53,303 @@ type CertificateRequestPolicyList struct {
...
-	// Allowed is the set of attributes that are "allowed" by this policy. A
-	// CertificateRequest will only be considered permissible for this policy if
-	// the CertificateRequest has the same or less as what is allowed.  Empty or
-	// `nil` allowed fields mean CertificateRequests are not allowed to have that
-	// field present to be permissible.
+	// Allowed defines the allowed attributes for a CertificateRequest.
+	// A CertificateRequest can request _less_ than what is allowed,
+	// but _not more_, i.e. a CertificateRequest can request a subset of what
+	// is declared as allowed by the policy.
+	// Omitted fields declares that the equivalent CertificateRequest
```
inteon, line 60:
```suggestion
	// Omitted fields declare that the equivalent CertificateRequest
```

inteon:
/approve
/lgtm
Thanks for doing this Erik and sorry that it turned out to be more work than expected.

erikgb:
> Thanks for doing this Erik and sorry that it turned out to be more work than expected.

No worries! I was expecting a marathon when starting on this. I am used to endless rounds of review on my suggested API extensions, and refining the API docs turned out to be even worse. 😅

PR #281, opened 2023-10-07
feat: add TODO for validation of fields with set semantics

erikgb:
While working on refining the CertificateRequestPolicy API docs in https://github.com/cert-manager/approver-policy/pull/280, I noticed that a few list-type fields are addressed as sets. This is something that the Kubernetes API can validate, and I think we should add validations that could help users of the API.

~These changes are potentially breaking for users, but I think it makes sense to introduce these additional validations in the current API version because:~

Update: We think this change is not worth potentially breaking the API for users, so I've just added TODO for the next API version for now.

/cc @inteon @SgtCoDFish


SgtCoDFish (commented):
I thought about this a bit, and I think I'm probably against introducing a breaking change for this.

While I do agree we should move fast with approver-policy and we shouldn't be scared to experiment and improve things, I think this change doesn't justify breaking policies which are working today.

I haven't tested, but my understanding from this PR and from the docs is that this would cause the following policy to break when it previously passed validation:

`dnsNames: { values: ["example.com", "test.example.com", "example.com"] }`

I think that kind of break would be really confusing to debug for users if it did break them.

I absolutely think we should set these as sets on v1alpha2  - that seems a no-brainer. But currently, somebody with the default approver disabled who's using approver-policy hasn't really got an easy alternative to use, and so breaking them like this doesn't outweigh the positives for me, personally.

I'd vote to add `TODO: add x-kubernetes-list-type: set in v1alpha2`  comments to each affected field, personally.

PR #290, opened 2023-10-21
refactor: support for multiple errors on same field

erikgb:
While working on https://github.com/cert-manager/approver-policy/pull/277, I realized that I need to be able to support multiple errors when evaluating each field on certificaterequests.

This PR is a simple refactoring from using `*field.Error` to `field.ErrorList` as the return type from evaluation functions. Also a minor improvement to one of the tests that were failing because of a bug in the refactoring. Is was impossible to spot the exact error without it. 😄 

/cc @inteon


inteon (approved):
/approve
/lgtm

PR #289, opened 2023-10-20
Metrics: add the `approved_count` and `unmatched_count` metrics

maelvls:
The only metrics approver-policy currently comes with is controller-runtime's. Although these metrics are useful to know whether the reconciliation loops are doing well, they are less than ideal to understand problems with misconfigured CertificateRequestPolicies.

The two use-cases are:

1. I'd like to know when there are "orphan" certificate requests, i.e., certificate requests that haven't been matched to any CertificateRequestPolicies.
2. I'd like to know when there is an increase in the number of denials. That might indicate a misconfiguration.

In this PR, I propose to add two metrics:

| Metric | Labels | Note | 
| --- | --- | --- |
| `approverpolicy_certificaterequest_orphan_count` (enum gauge) | `namespace`\* | 0-values not shown | 
| `approverpolicy_certificaterequest_approved_count` (gauge) | `namespace`, `approved_status` (False, True) | 0-values not shown |

By "0-values not shown", I mean that the metric only shows a value when it has a non-0 value. Since the metric is labelled by namespace, you will only see the namespaces that have at least one orphan for the `orphan_count` metric (same goes with the `approved_count` metric). We haven't implemented the extra `orphan_count_total` and  `approved_count_total` as part of this PR because I didn't know that `count_total` was meant to circumvent this 0-value limitation.

We haven't added the label `name` for `unmatched_count` because:
-  It may end up being very big (cardinality problem) in case of
many CRs being created at the same time and if the metrics are scrapped
right before the approver has had time to set the Approved condition.
- The `name` isn't often useful, particularly with the csi-driver. It would be much more useful to have a label `owner_name` and `owner_kind`.

**Testing it yourself:**

First, install cert-manager without its built-in approver:

```bash
helm upgrade --install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --set installCRDs=true \
  --set prometheus.enabled=true \
  --set prometheus.servicemonitor.enabled=true \
  --set prometheus.servicemonitor.labels.instance=primary \
  --set extraArgs="{--controllers=*\,-certificaterequests-approver,--feature-gates=StableCertificateRequestName=true}"
```

Then, install the Grafana Agent Operator:

```bash
helm repo add --force grafana https://grafana.github.io/helm-charts
helm upgrade --install -n grafana --create-namespace grafana-agent-operator grafana/grafana-agent-operator
```

Then, go to https://grafana.com/. Create an account, and find the credentials to the Prometheus push endpoint:

grafana-cloud-1

[...]

PR #293, opened 2023-10-27
Advanced Metrics: Add the "denied_count" metric to account for the Denied=True condition

maelvls:
I made a big mistake in #289. I mistakenly thought `Approved=False` would be reported when an approver denies a CR. But this morning, I double checked and found out that "denial" isn't reported with the `Approved` condition! A denied CR status looks like this:

```yaml
 - type: Denied
   status: "True"
   reason: policy.cert-manager.io
   message: 'No policy approved this request: [issuer-2: spec.allowed.dnsNames.values:
     Invalid value: []string{"forbidden-domain-41.com"}: *.example.com, *.ca-wont-accept.org]'
```

I hadn't properly read the conditions' documentation:

> "Approved" indicates that a certificate request is approved and ready for signing. Condition must never have a status of `False`, and cannot be modified once set. Cannot be set alongside `Denied`.

and

> "Denied" indicates that a certificate request is denied, and must never be signed. Condition must never have a status of `False`, and cannot be modified once set. Cannot be set alongside `Approved`.

In this PR, I introduce a third metrics "denied_count" to fix the issue. I also removed the `approved_status` since it doesn't make sense anymore. The metrics are now:

| Metric | Labels |
|--|--|
| `approverpolicy_certificaterequest_denied_count` (gauge) | `namespace` |  
| `approverpolicy_certificaterequest_approved_count` (gauge) | `namespace`* | 
| `approverpolicy_certificaterequest_unmatched_count` (gauge) | `namespace` | 

*Before this PR, "approved_count" had a label `approved_status` (False, True). Since "denied_count" already captures this information, I have removed the label `approved_status`.

### DIFF
--- a/pkg/internal/metrics/metrics.go
+++ b/pkg/internal/metrics/metrics.go
@@ -27,10 +27,26 @@ var (
 	// get removed over time e.g. with revisionHistoryLimit.
 	approvedCount = prometheus.NewDesc(
 		"approverpolicy_certificaterequest_approved_count",
-		"Number of CertificateRequests that were been approved or denied (Approved=True or Approved=False).",
+		"Number of CertificateRequests that have been approved (Approved=True).",
 		[]string{
 			"namespace",
-			"approved_status",
+		},
+		nil,
+	)
+
+	// deniedCount counts the number of CertificateRequest currently denied by
+	// looking at the Denied condition.
+	//
+	// - type: Denied
+	//   status: "True"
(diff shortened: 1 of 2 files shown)


inteon:
This PR makes a lot of sense.
We haven't released this yet, so not really a "big mistake".
For some reason I also missed this when reviewing the original PR 😅.
/approve
/lgtm

PR #295, opened 2023-10-29
refactor: prepare code for introducing CEL validations

erikgb:
I am still waiting for reviews on https://github.com/cert-manager/approver-policy/pull/277, and I think one of the reasons is that it appears big (a lot of changes). I know the feeling, so here is a preparation PR with some required refactorings:

- Always use the constructor to create `allowed` - since I want to init the upcoming CEL expression cache in this constructor. I have done the same for `constraints` for consistency.
- Introduced an interface to move the basic eval functions to `allowed`. This is required to get transparent access to the CEL expression cache singleton.

/cc @inteon

### DIFF
--- a/pkg/internal/approver/allowed/allowed.go
+++ b/pkg/internal/approver/allowed/allowed.go
@@ -30,7 +30,7 @@ import (
 
 // Load the allowed approver.
 func init() {
-	registry.Shared.Store(allowed{})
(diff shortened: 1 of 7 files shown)


erikgb:
Dropping this PR. Suggesting to continue in https://github.com/cert-manager/approver-policy/pull/277. Hope to get some progress there....

PR #298, opened 2023-11-02
Use cert-manager's servertls.CertificateSource and simplify webhook code

inteon:
We were using cert-manager's authority.DynamicAuthority to issue the webhook certificate, but implemented our own custom wrapper here. Much simpler is to just use the upstream servertls.CertificateSource wrappers instead.


pkg/internal/cmd/cmd.go
```diff
@@ -70,25 +98,31 @@ func NewCommand(ctx context.Context) *cobra.Command {
 					BindAddress: opts.MetricsAddress,
 				},
 				WebhookServer: ctrlwebhook.NewServer(ctrlwebhook.Options{
-					Port:    opts.Webhook.Port,
-					Host:    opts.Webhook.Host,
```
erikgb, line 74: It seems like this option is no longer used, or? Maybe clean up the Helm chart if that's the case?

inteon: Sorry, I missed this. Also, I missed some other parts, the PR should be ready now.

erikgb (approved):
I don't know how the cert-manager `CertificateSource` works, but tests pass and code is removed.  👍 Maybe add a to-be release note about the certificate directory no longer possible (and needed) to configure?

/lgtm

inteon:
As far as I understand, the certDir was only used as an implementation detail (where to store the dynamically generated certificates).
In the future we might want to support a new option that lets you use a mounted certificate instead of a dynamically generated certificate.
Also, maybe we should also add support for dynamically generated certificates to trust-manager.

/approve

Issue #301, opened 2023-11-07
Typo in error message: connection patch should say CertificateRequest.Status patch


wallrj:
In https://github.com/cert-manager/approver-policy/issues/297 I noticed a typo in the error message that @erikgb reported:

> failed to apply connection patch: admission webhook \"webhook.cert-manager.io\" denied the request

Should say "failed to apply CertificateRequest.Status patch.

The word "connection" is also misused elsewhere:

 * https://github.com/search?q=repo%3Acert-manager%2Fapprover-policy%20connection%20patch&type=code

_Originally posted by @wallrj in https://github.com/cert-manager/approver-policy/pull/186#discussion_r1385027477_

PR #302, opened 2023-11-07
fix: typos in error message wrapping

erikgb:
Fixes https://github.com/cert-manager/approver-policy/issues/301

/cc @wallrj


wallrj (approved):
Thanks!

/approve
/lgtm
/hold in case you also want to fix the `con` variable name. Else unhold and do that in another PR.

erikgb:
> in case you also want to fix the con variable name

Thanks! Good catch. Done.

wallrj (approved):
/lgtm

Issue #297, opened 2023-10-31
Attempt to update status.conditions denied by cert-manager webhook


erikgb:
We have been running cert-manager for a long time in multi-tenant clusters - with the default built-in approver. Yesterday we promoted approver-policy to one of our busiest cluster (TEST). While everything seems to work as it should: new certificate requests are approved by approver-policy, we notice significant log noise in approver policy logs. Here is an example:

````
E1031 08:40:15.331193 1 controller.go:329] controller-manager "msg"="Reconciler error" "error"="failed to apply connection patch: admission webhook \"webhook.cert-manager.io\" denied the request: status.conditions: Forbidden: 'Approved' condition may not be modified once set" "CertificateRequest"={"name":"switchstatus-adapter-tls-53","namespace":"int-switchstatus"} "controller"="certificaterequest" "controllerGroup"="cert-manager.io" "controllerKind"="CertificateRequest" "name"="switchstatus-adapter-tls-53" "namespace"="int-switchstatus" "reconcileID"="00c940e8-87d9-4ae9-ae46-5c258203e7af"
````

CertificateRequest (after the error is logged):

````yaml
apiVersion: cert-manager.io/v1
kind: CertificateRequest
metadata:
  annotations:
    cert-manager.io/certificate-name: switchstatus-adapter-tls
    cert-manager.io/certificate-revision: '53'
    cert-manager.io/private-key-secret-name: switchstatus-adapter-tls-z7w74
  resourceVersion: '3060778645'
  name: switchstatus-adapter-tls-53
  uid: b93e2a35-89bb-4186-9332-3c5e971eba0a
  creationTimestamp: '2023-10-30T13:44:46Z'
  generation: 1
  managedFields:
    - apiVersion: cert-manager.io/v1
      fieldsType: FieldsV1
      fieldsV1:
        'f:status':
          'f:conditions':
            'k:{"type":"Approved"}':
              .: {}
              'f:lastTransitionTime': {}
              'f:message': {}
              'f:reason': {}
              'f:status': {}
              'f:type': {}
      manager: approver-policy
      operation: Apply
      subresource: status
      time: '2023-10-30T13:45:09Z'
    - apiVersion: cert-manager.io/v1
      fieldsType: FieldsV1
      fieldsV1:
        'f:status':
          'f:ca': {}
          'f:certificate': {}
          'f:conditions':
            'k:{"type":"Approved"}':
              .: {}
              'f:lastTransitionTime': {}
              'f:message': {}
[...]

PR #303, opened 2023-11-09
fix: should not change existing condition of type

erikgb:
When attempting to implement one of the suggested solutions to https://github.com/cert-manager/approver-policy/issues/297, I discovered that there is a bug in the code: we don't take existing certificaterequest conditions into account when preparing the SSA patch.

IMO we could implement SSA with code that is easier to read and understand. Which could reduce the chance of bugs like this one. Hopefully, we can discuss how to get some progress on https://github.com/cert-manager/trust-manager/pull/217. Since I started in trust-manager I suggest you give me a chance to show you which improvements I have in mind. If it's successful, I am happy to do a similar effort in this project.

Fixes https://github.com/cert-manager/approver-policy/issues/297 (hopefully, if not I will just reopen it)

/cc @wallrj @inteon

### DIFF
--- a/pkg/internal/controllers/certificaterequests.go
+++ b/pkg/internal/controllers/certificaterequests.go
@@ -176,6 +176,11 @@ func (c *certificaterequests) reconcileStatusPatch(ctx context.Context, req ctrl
 		return ctrl.Result{}, nil, client.IgnoreNotFound(err)
 	}
 
+	if apiutil.CertificateRequestIsApproved(cr) || apiutil.CertificateRequestIsDenied(cr) {
+		// Return early if already approved/denied as this is decision is final for requests.
+		return ctrl.Result{}, nil, nil
+	}
+
 	// Query review on the approver manager.
 	response, err := c.manager.Review(ctx, cr)
 	if err != nil {
--- a/pkg/internal/controllers/certificaterequests.go
+++ b/pkg/internal/controllers/certificaterequests.go
@@ -26,12 +26,14 @@ import (
 	apiutil "github.com/cert-manager/cert-manager/pkg/api/util"
 	cmapi "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	cmmeta "github.com/cert-manager/cert-manager/pkg/apis/meta/v1"
+	"github.com/cert-manager/issuer-lib/conditions"
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/clock"
 	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -52,6 +54,9 @@ import (
 type certificaterequests struct {
 	// log is logger for the certificaterequests controller.
 	log logr.Logger
+
+	// clock returns time which can be overwritten for testing.
+	clock clock.Clock
 
 	// recorder is used for creating Kubernetes events on resources.
 	recorder record.EventRecorder
@@ -76,6 +81,7 @@ type certificaterequests struct {
 func addCertificateRequestController(ctx context.Context, opts Options) error {
 	c := &certificaterequests{
 		log:      opts.Log.WithName("certificaterequests"),
+		clock:    clock.RealClock{},
 		recorder: opts.Manager.GetEventRecorderFor("policy.cert-manager.io"),
 		client:   opts.Manager.GetClient(),
 		lister:   opts.Manager.GetCache(),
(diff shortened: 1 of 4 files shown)


pkg/internal/controllers/certificaterequests.go
```diff
@@ -195,8 +195,8 @@ func (c *certificaterequests) reconcileStatusPatch(ctx context.Context, req ctrl
 		log.V(2).Info("approving request")
 		c.recorder.Event(cr, corev1.EventTypeNormal, "Approved", response.Message)
```
inteon, line 197: Could we instead just return without doing a patch?

erikgb: I would prefer to not do it in this PR. That will require a lot of additional changes/refactoring. 

Not a bad idea, it will only require rewriting code and test.

erikgb:
> > Fixes [cert-manager/trust-manager#217](https://github.com/cert-manager/trust-manager/pull/217) (hopefully, if not I will just reopen it)
> 
> Is that link correct? I wouldn't expect this to fix a trust-manager PR 😁

Whooops, thanks! Fixed now.

pkg/internal/controllers/certificaterequests.go
erikgb: It seems like I was a bit quick to reject your suggestion, @inteon. After a walkthrough of this PR with @wallrj in today's stand-up, I realize that your suggestion makes a lot of sense to do now, and here it is: https://github.com/cert-manager/approver-policy/pull/303/commits/bb75196b1e98dcb078eb70888795c95529cfdde6. I still think we should bugfix the conditions patch - just in case the controller cache is stale.

erikgb:
/hold

I would like to squash/rebase this PR before merging. But please review. 😉

pkg/internal/controllers/certificaterequests.go
```diff
@@ -195,8 +200,8 @@ func (c *certificaterequests) reconcileStatusPatch(ctx context.Context, req ctrl
...
 
-		c.setCertificateRequestStatusCondition(
-			&crPatch.Conditions,
+		crPatch.Conditions = c.getCertificateRequestStatusConditions(
```
inteon, line 203: @erikgb do we still need these changes?

inteon: I think the lines 179-183 do fix the issue already right?

erikgb: Yes, either will fix issues. This is a bug IMO - regardless of the early exit.

erikgb: The early exit just hides the bug. But if the cache is stale, we will be hit if not fixing this also.

pkg/internal/controllers/certificaterequests.go
```diff
@@ -238,22 +243,14 @@ func (c *certificaterequests) reconcileStatusPatch(ctx context.Context, req ctrl
 // Update the status with the provided condition details & return
 // the added condition.
 // NOTE: this code is just a workaround for apiutil only accepting the certificaterequest object
-func (c *certificaterequests) setCertificateRequestStatusCondition(
-	conditions *[]cmapi.CertificateRequestCondition,
+func (c *certificaterequests) getCertificateRequestStatusConditions(
```
inteon, line 246: Could we use the https://github.com/cert-manager/issuer-lib/blob/main/conditions/certificaterequest.go implementation instead?

inteon: We fixed these issues in that repo (see eg. https://github.com/cert-manager/issuer-lib/commit/aa4880ccbbe50d10992fb103aada6402694745b2).

erikgb: I'll give it a try.

pkg/internal/controllers/certificaterequests.go
erikgb: I also noticed this now "duplicated" line:

https://github.com/cert-manager/approver-policy/blob/c8f950602e17a0a6fc0d0ef78dd61db20b387d59/pkg/internal/controllers/certificaterequests.go#L100

This should probably be cleaned up. And makes me even more sure we must have the `lastTransitionTime` bugfix.

pkg/internal/controllers/certificaterequests.go
erikgb: It's a new dependency, and I am not a big fan of dependencies. 😉 But if we think this new dependency is part of the future of this project it seems to work. Incoming commit.

erikgb: I stayed away from the latest version to avoid being forced to upgrade to Go 1.21.

erikgb: Please review and let me know what you think. I want to rebase before we merge something. Keeping the commits until we decide how to fix the issue.

inteon: Sorry for the ambiguity, I was thinking we could copy the code from issuer-lib. I don't think we should have it as a dependency, since we are not building an issuer. What do you think?

erikgb: Ok, works for me. 😄 I want to get this issue fixed. After discovering the even earlier exit, I suggest to leave that duplication out and focus on the bug here: not taking `lastTranstionTime` from and existing condition into account. I wrote a test just for that - which is a bit pointless as the PR stands now. Since we exit before it becomes relevant.

Personally, I think the code in issuer-lib is too complex and hard to read. But I really want this bug fixed.

erikgb: Copied the function from issuer-lib now.

pkg/internal/controllers/certificaterequests.go
erikgb: I suggest we now remove the duplicated early exit, ref.

https://github.com/cert-manager/approver-policy/blob/c8f950602e17a0a6fc0d0ef78dd61db20b387d59/pkg/internal/controllers/certificaterequests.go#L115-L118

Isn't a predicate better and sufficient?

inteon (approved):
Thank you for fixing this bug!
/approve
/lgtm

erikgb:
@inteon Should I rebase before merge?

inteon:
@erikgb sorry I thought it had been rebased already, feel free to create a new PR to fix any remaining problems.

PR #305, opened 2023-11-14
fix: dynamic cert runnable must NOT be leader election aware

erikgb:
In https://github.com/cert-manager/approver-policy/pull/298 we unintentionally made the dynamic cert runnable leader election aware - which is the default for runnables in controller-runtime.

Since all replicas serve webhook endpoints, the runnable providing webhook certificates must be allowed to run without leader election.

This was probably not caught by our tests since we run only a single replica. And then this is not an issue.

### DIFF
--- a/pkg/internal/cmd/cmd.go
+++ b/pkg/internal/cmd/cmd.go
@@ -100,9 +100,7 @@ func NewCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("unable to create controller manager: %w", err)
 			}
 
-			if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
-				return certificateSource.Run(ctx)
-			})); err != nil {
+			if err := mgr.Add(&dynamicCertRunnable{source: certificateSource}); err != nil {
 				return err
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/internal/cmd/cmd.go
```diff
@@ -143,3 +141,19 @@ func NewCommand(ctx context.Context) *cobra.Command {
 
 	return cmd
 }
+
+type dynamicCertRunnable struct {
```
inteon, line 145: Could we make this something generic? So we can re-use it in cert-manager (I think I made the same mistake there).

erikgb: I am not sure if that's possible, but I am not familiar with Golang generics. But I know that generics is not supported for methods, only functions.

erikgb: But feel free to open another PR to fix this. My main concern is to fix the bug.

inteon: I meant making it generic by letting it accept a function as argument (not using actual generics).

inteon (commented):
/approve
/lgtm

inteon:
Thanks for fixing the bug @erikgb

pkg/internal/cmd/cmd.go
erikgb: Ahhh, sorry for being a bit slow. Can we fix this by making the `DynamicSource` implement these interfaces? That will make it easier to use across projects.

PR #309, opened 2023-11-21
Don't use ROOTDIR environment variable in tests.

inteon:
We don't want to code the relative paths for test assets in Go.
Instead, we should configure that in our Make files and pass them as separate environment variables to Go.

### DIFF
--- a/pkg/internal/approver/manager/review_test.go
+++ b/pkg/internal/approver/manager/review_test.go
@@ -31,7 +31,7 @@ import (
 	"github.com/cert-manager/approver-policy/pkg/approver/fake"
 	"github.com/cert-manager/approver-policy/pkg/approver/manager"
 	"github.com/cert-manager/approver-policy/pkg/internal/approver/manager/predicate"
-	"github.com/cert-manager/approver-policy/test/env"
+	testenv "github.com/cert-manager/approver-policy/test/env"
 )
 
 func Test_Review(t *testing.T) {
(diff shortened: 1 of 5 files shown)


test/env/util.go
```diff
@@ -53,14 +53,14 @@ users:
...
 )
 
-// RootDirOrSkip returns the ROOTDIR environment variable, or skips if the
+// VariableOrFail returns the specified environment variable, or fails if the
 // variable is undefined or empty.
-func RootDirOrSkip(t *testing.T) string {
-	rootDir := os.Getenv("ROOTDIR")
-	if len(rootDir) == 0 {
-		t.Skip("WARNING: skipping test as 'ROOTDIR' is not defined")
+func VariableOrFail(t *testing.T, name string) string {
```
erikgb, line 58:
```suggestion
func GetenvOrFail(t *testing.T, name string) string {
```

I got a bit confused by this function name. The docs (and code) were clarifying, but I think the func name could be improved.

pkg/internal/approver/manager/review_test.go
```diff
@@ -41,10 +40,9 @@ func Test_Review(t *testing.T) {
 		cancel()
 	})
 
-	rootDir := env.RootDirOrSkip(t)
 	env := env.RunControlPlane(t, ctx,
-		filepath.Join(rootDir, "_bin/cert-manager"),
-		filepath.Join(rootDir, "deploy/charts/approver-policy/templates/crds"),
+		env.GetenvOrFail(t, "CERT_MANAGER_CRDS"),
+		env.GetenvOrFail(t, "APPROVER_POLICY_CRDS"),
```
erikgb, line 45: Maybe use a consistent import alias, `testenv`? Or will that collide with https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.16.3/pkg/envtest? No, I think it will work fine.

inteon: done

erikgb (approved):
/lgtm

Thanks!

erikgb:
> crd.go:82: lstat /home/prow/go/src/github.com/cert-manager/approver-policy//home/prow/go/src/github.com/cert-manager/approver-policy/_bin/cert-manager/crds.yaml: no such file or directory

There is a bug in a path somewhere.

Issue #216, opened 2023-03-23
Simplify configuration by creating RBAC by default


inteon:
Currently, it is quite complex to configure approver-policy due to all the necessary additional RBAC.

The following RBAC could be added to the Helm chart to simplify usage (easy-mode):
1. allow approver-policy to approve all issuer types:
```yaml
- apiGroups: ["cert-manager.io"]
  resources: ["signers"]
  verbs: ["approve"]
...
- kind: ServiceAccount
  name: {{ include "cert-manager-approver-policy.name" . }}
  namespace: {{ .Release.Namespace }}
```

2. make all policies applicable to the cert-manager SA by default (use selector for filtering instead):
```yaml
- apiGroups: ["policy.cert-manager.io"]
  resources: ["certificaterequestpolicies"]
  verbs: ["use"]
...
- kind: ServiceAccount
  name: cert-manager
  namespace: cert-manager
```

erikgb:
It seems like 1. was fixed by https://github.com/cert-manager/approver-policy/pull/416, but as far as I can see 2. still remains unresolved.

/help

PR #310, opened 2023-11-25
docs: brush-up policy examples

erikgb:
While considering a doc update PR in our website project to add some docs around the new CEL validations feature, I noticed that our docs link to policy examples in this repo. We should have some examples showing how the new CEL validation feature can be used, so I updated the all-options example to really use all options.

In addition, I removed the repeated RBAC setup for all examples. IMO the examples here should be policy examples.

I also removed the last occurrence of the slightly confusing term "impossible condition" in this project. 😉


docs/examples/all-options.yaml
```diff
@@ -7,79 +8,103 @@ spec:
...
+      - "server auth"
+      - "client auth"
     subject:
       organizations:
-        values: ["hello-world"]
+        required: false
+        values: ["*"]
+        validations:
+          - rule: self.size() > 0
+            message: must not be empty
```
inteon, line 52: @erikgb is this a good example? We say the field is not required but still require it to be not empty.

docs/examples/all-options.yaml
```diff
@@ -7,79 +8,103 @@ spec:
...
   constraints:
     minDuration: 1h
     maxDuration: 24h
     privateKey:
       algorithm: RSA
       minSize: 2048
       maxSize: 4096
-
+  plugins:
+    rego:
```
inteon, line 103: This plugin is not available in the default distribution.

docs/examples/example.com.yaml
```diff
@@ -8,46 +8,19 @@ spec:
...
       algorithm: RSA
       minSize: 2048
-
   selector:
     issuerRef:
       name: letsencrypt-prod
       kind: Issuer
       group: cert-manager.io
----
-kind: Role
```
inteon, line 27: Is there a reason we remove this RBAC?

docs/examples/all-options.yaml
erikgb: I agree it's not a good example, but I was a bit tired of coming up with good examples. 😅 The idea was to validate that the subject attributes should not be allowed to set to the empty string (`""`), but I think that's already covered by `required`. Do you have any good suggestions?

docs/examples/all-options.yaml
erikgb: I know, and no plugins are. This policy example is a all-options/max example, so I think we should include set values on all fields available in the API.

docs/examples/example.com.yaml
erikgb: I wrote about this in the PR description. To me, it makes little sense to repeat the repetitive RBAC in the policy examples. They are not part of the policy, but required to make the policy effective. In our setup we have granted cert-manager SA RBAC to all policies, by not setting `resourceNames`. I think that is common setup, but I might be wrong. I'm ok to leave the RBAC as is if that's a requirement to get this merged. Let me know what you think! 😄

inteon: It makes sense. We just have to make sure we clearly document what RBAC is required (somewhere else).

erikgb: It is documented [here](https://cert-manager.io/docs/policy/approval/approver-policy/#configuration). Any documentation can be improved, but at least I could understand it when configuring our approver-policy installation.

erikgb: If we implemented https://github.com/cert-manager/approver-policy/issues/216, this wouldn't be something to discuss. 😉

docs/examples/all-options.yaml
inteon: I tried to fix it by supplying an empty array: https://github.com/cert-manager/approver-policy/pull/310/commits/87762ac2dc3ae7dea34b313d745cc0342db12a95

erikgb: Nice! I like it! It shows that the field is available, which I think is the main point of this "example".

PR #311, opened 2023-11-25
chore(deps): upgrade github.com/google/cel-go

erikgb:
It seems like Depandabot needs a little help in https://github.com/cert-manager/approver-policy/pull/307. Our tests are fragile with this dependency to a fast-moving upstream project. I will try to improve this in a follow-up PR.

/cc @inteon


inteon:
Thanks for fixing this.
/approve
/lgtm

PR #313, opened 2023-11-27
Metrics: add the metric `approverpolicy_certificaterequestpolicy_status`

maelvls:
On its own, the approver-policy project doesn't really need a metric about the statuses of CertificateRequestPolicies since the status will always be `Ready=True` and the status always looks like this:

```yaml
status:
  conditions:
    - type: Ready
      reason: Ready
      status: "True"
      message: CertificateRequestPolicy is ready for approval evaluation
```

This metric becomes useful when implementing extensions to the approver-policy
project using the `spec.plugin` field. In that case, the plugin may report
errors, and I would like to propose a new metric to be alerted of malfunctioning
CertificateRequestPolicies.

The metric labels are:

| Label | Description |
| ----- | ----------- |
| `ready_status` | `True` or `False`. That's the status of the CertificateRequestPolicy's Ready condition. The `Unknown` status is not used by the project. |
| `namespace` | The namespace of the CertificateRequestPolicy. |
| `name` | The name of the CertificateRequestPolicy. |

Example when scraping `/metrics`:

```text
approverpolicy_certificaterequestpolicy_status{ready_status="True",namespace="default",name="policy1"} 1
approverpolicy_certificaterequestpolicy_status{ready_status="False",namespace="default",name="policy2"} 1
```

## Why are `reason` or `message` not included in the metric?

The `reason` field because it is paraphrasing `status` (the `reason` says
"Ready" or "NotReady"). That's why it isn't included in the metric.

We don't report the `message` field from the condition because the `message`
field is a free-text string by definition. It can contain changing values
(timestamps, names of resources) that may change for each object. This makes it
impractical for Prometheus due to the cardinality explosion.

## Testing it Yourself

You can find the instructions to set up this PR with Grafana and Prometheus [here](https://github.com/cert-manager/approver-policy/pull/289).


pkg/internal/metrics/metrics.go
```diff
@@ -79,6 +81,23 @@ var (
...
+	// Ready condition. We don't report the `message` field from the condition
+	// because it isn't machine-readable (i.e., it's a free-text string that
+	// changes for each object and isn't meant to be used for filtering and
+	// grouping in Prometheus). We also don't report the `reason` field because
+	// it is paraphrasing the `status` (it says "Ready" or "NotReady").
+	crPolicyStatus = prometheus.NewDesc(
+		"approverpolicy_certificaterequestpolicy_status",
+		"Status of the CertificateRequestPolicy resources.",
+		[]string{
+			"namespace",
```
erikgb, line 95: Why do we include a namespace label? `CertificateRequestPolicy` is a cluster-scoped resource AFAIK.

maelvls: I am stupid. This namespace label is totally useless 😅 I’ll remove it.

erikgb: @maelvls Don't be so hard on yourself!

maelvls: Done!

go.mod
```diff
@@ -93,7 +93,7 @@ require (
 	golang.org/x/net v0.17.0 // indirect
 	golang.org/x/oauth2 v0.13.0 // indirect
 	golang.org/x/sync v0.4.0 // indirect
-	golang.org/x/sys v0.14.0 // indirect
+	golang.org/x/sys v0.13.0 // indirect
```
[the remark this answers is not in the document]
maelvls: My mistake, I am unsure how I got this `go.mod` change. I should probably have checked before pushing 😇

erikgb:
@maelvls I am looking at some equivalent metrics produced by kube-state.metrics. Example:

````
kube_pod_status_ready{condition="false", container="kube-state-metrics", endpoint="http", instance="10.101.20.20:8080", job="kube-state-metrics", namespace="accurate", pod="accurate-controller-manager-7d6cdc897b-28hsq", prometheus="monitoring/k8s", service="monitoring-kube-state-metrics", uid="7597aca9-9a6e-4e4a-a516-d300a102299f"}	0
kube_pod_status_ready{condition="true", container="kube-state-metrics", endpoint="http", instance="10.101.20.20:8080", job="kube-state-metrics", namespace="systemcatalog-operator", pod="systemcatalog-operator-controller-manager-7c6b4b85fb-rr7jm", prometheus="monitoring/k8s", service="monitoring-kube-state-metrics", uid="35157dc7-e371-49cb-87f7-6ae81e538086"}	1
kube_pod_status_ready{condition="unknown", container="kube-state-metrics", endpoint="http", instance="10.101.20.20:8080", job="kube-state-metrics", namespace="collectorforopenshift", pod="collectorforopenshift-8pm95", prometheus="monitoring/k8s", service="monitoring-kube-state-metrics", uid="e31e7614-dd24-4010-b42b-0a8cfacc3afe"}	0
````

I think we should build something equivalent. When comparing with your proposal, I think we should:
- rename metric to `approverpolicy_certificaterequestpolicy_ready`
- rename the `ready_status` label to `condition` and lowercase the label value (Prom best-practice anyway)
- make the metric value `1` only if the policy if condition `Ready=true`, `0` otherwise

WDYT? Another solution is just to use kube-state-metrics to produce these metrics, ref. https://github.com/cert-manager/cert-manager/pull/6415#issuecomment-1824441005

PR #316, opened 2023-12-13
Add Helm option to not include CRD

inteon:
TLDR; Adds the option to not install the CRD as part of the Helm chart. The CRD must still be installed separately.

If you install the approver-policy Helm chart with the `crds.enabled: true` option, uninstalling the Helm chart will result in also uninstalling the CRDs. This will trigger the CRs to be garbage collected and deleted too.
To prevent uninstalling the CRDs when the Helm chart is uninstalled, the CRDs can be installed separately. With the change proposed in this PR, this can be done as follows:

```bash
$ helm template cert-manager-approver-policy jetstack/cert-manager-approver-policy | \
        yq e '. | select(.kind == "CustomResourceDefinition")' \
       > approver-policy-crds.yaml

$ kubectl apply --server-side -f approver-policy-crds.yaml

$ helm upgrade cert-manager-approver-policy jetstack/cert-manager-approver-policy \
        -i -n cert-manager \
        --wait \
        --set crds.enabled=false
```

See https://github.com/cert-manager/trust-manager/pull/102#issuecomment-1400131204 for the equivalent change in trust-manager.

### DIFF
--- a/deploy/charts/approver-policy/values.yaml
+++ b/deploy/charts/approver-policy/values.yaml
@@ -109,3 +109,6 @@ commonLabels: {}
 # -- Optional allow custom annotations to be placed on cert-manager-approver pod
 podAnnotations: {}
 
+crds:
+  # -- Whether or not to install the crds.
+  enabled: true
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
-deploy/charts/approver-policy/templates/crds/* linguist-generated=true
+deploy/charts/approver-policy/templates/crd-* linguist-generated=true
 deploy/charts/approver-policy/README.md linguist-generated=true
 docs/api/api.md linguist-generated=true
 
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,7 @@ HELM_VERSION ?= 3.11.3
 KUBEBUILDER_TOOLS_VERSION ?= 1.27.1
 # Check https://github.com/kyverno/kyverno/releases for latest available release
 KYVERNO_VERSION ?= v1.10.0
+YQ_VERSION ?= v4.34.1
 K8S_CLUSTER_NAME ?= approver-policy
 IMAGE_REGISTRY ?= quay.io/jetstack
 IMAGE_NAME := cert-manager-approver-policy
@@ -91,8 +92,18 @@ image: ## build docker image
 .PHONY: generate-manifests
 generate-manifests: ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 generate-manifests: | $(BINDIR)/controller-gen
+	$(eval crds_gen_temp := $(BINDIR)/scratch/crds)
+
 	$(BINDIR)/controller-gen rbac:roleName=manager-role crd webhook paths="./..." \
-		output:crd:artifacts:config=$(helm_chart_source_dir)/templates/crds
+		output:crd:artifacts:config=$(crds_gen_temp)
+
+	echo "Updating CRDs with helm templating, writing to $(helm_chart_source_dir)/templates"
+
+	for i in $$(ls $(crds_gen_temp)); do \
+	echo "{{ if .Values.crds.enabled }}" > $(helm_chart_source_dir)/templates/crd-$$i; \
+	cat $(crds_gen_temp)/$$i >> $(helm_chart_source_dir)/templates/crd-$$i; \
+	echo "{{ end }}" >> $(helm_chart_source_dir)/templates/crd-$$i; \
+	done
 
 .PHONY: generate-deepcopy
 generate-deepcopy: ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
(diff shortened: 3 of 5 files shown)


deploy/charts/approver-policy/values.yaml
```diff
@@ -109,3 +109,6 @@ commonLabels: {}
 # -- Optional allow custom annotations to be placed on cert-manager-approver pod
 podAnnotations: {}
 
+crds:
+  # -- Whether or not to install the crds.
+  enabled: true
```
erikgb, line 114: Any reason for using a value different from the one used in cert-manager, ref. https://cert-manager.io/docs/installation/helm/#helm-installation?

inteon: It's based on trust-manager: https://github.com/cert-manager/trust-manager/blob/main/deploy/charts/trust-manager/values.yaml#L184-L186

erikgb: Thanks, makes sense!

erikgb (commented):
I think

https://github.com/cert-manager/approver-policy/blob/c459a137efbbe435f3033a3aa751b63da8cbc552/Makefile#L181

must be updated. But why are we changing the location of the CRD? Can we Helm template it in place?

inteon:
I moved the crd because it was very confusing (instead of placing the crds in `templates/` or `crds/` it was placed in `templates/crds/`).

wallrj (changes requested):
@inteon For the benefit of the changelog in the next release, please write *why* you're adding this option. Not being a pain, I honestly can't remember. What problem does it solve?

The documentation in the chart README does not say anything about why you might want to omit the CRDs. And the https://cert-manager.io/docs/policy/approval/approver-policy/ document doesn't talk about it either.

You mentioned that the option is based on a similar option in trust-manager, but I can't find any reason for the option there either:

```
$ helm inspect readme jetstack/trust-manager | grep crds
| crds.enabled | bool | `true` | Whether or not to install the crds. |
```

inteon:
@wallrj same reason as https://github.com/cert-manager/trust-manager/pull/102

wallrj:
> @wallrj same reason as [cert-manager/trust-manager#102 (comment)](https://github.com/cert-manager/trust-manager/pull/102#issuecomment-1400131204)

Ok. Then put something in the PR title and description so that whoever next releases approver-policy knows what to put in the release notes.

Here's my suggestion:
> Platform engineer needs to uninstall approver-policy without deleting all the CertificateRequestPolicy resources in the cluster. For example, when migrating from a vanilla approver-policy to an approver-policy "plugin", such as approver-policy-enterprise.
> In this PR we have partially solved this problem by allowing the approver-policy chart be installed without the CRDS,
> using the new `crds.enabled: false` value. This way the the platform engineer can uninstall the vanilla approver-policy chart and then install an approver-policy plugin chart without interfering with the CertificateRequestPolicy resources in the cluster.
> 
> How to extract the CRDs from the chart and install them manually, is currently currently left as an exercise for the platform engineer.
> In a follow up PR, in https://github.com/cert-manager/makefile-modules, we will implement a new `make release` feature to publish the CRDs to a separate downloadable manifest file attached to the GitHub release.
> We will also add documentation explaining why and how production ready deployments should install the CRDs before installing the chart.

inteon:
> > @wallrj same reason as [cert-manager/trust-manager#102 (comment)](https://github.com/cert-manager/trust-manager/pull/102#issuecomment-1400131204)
> 
> Ok. Then put something in the PR title and description so that whoever next releases approver-policy knows what to put in the release notes.
> 
> Here's my suggestion:
> 
> > Platform engineer needs to uninstall approver-policy without deleting all the CertificateRequestPolicy resources in the cluster. For example, when migrating from a vanilla approver-policy to an approver-policy "plugin", such as approver-policy-enterprise.
> > In this PR we have partially solved this problem by allowing the approver-policy chart be installed without the CRDS,
> > using the new `crds.enabled: false` value. This way the the platform engineer can uninstall the vanilla approver-policy chart and then install an approver-policy plugin chart without interfering with the CertificateRequestPolicy resources in the cluster.
> > How to extract the CRDs from the chart and install them manually, is currently currently left as an exercise for the platform engineer.
> > In a follow up PR, in https://github.com/cert-manager/makefile-modules, we will implement a new `make release` feature to publish the CRDs to a separate downloadable manifest file attached to the GitHub release.
> > We will also add documentation explaining why and how production ready deployments should install the CRDs before installing the chart.

I updated the PR details with the exact instructions necessary to install the CRDs separately and why you would like to do this.

wallrj:
I think it was you who told me about `"helm.sh/resource-policy": keep` in [Tell Helm Not To Uninstall a Resource](https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource); why not do that instead?

If the instructions are to use `yq` to get only the CRDs, then why not solve this problem by also using `yq` to get everything **but** the CRDs?
```sh
helm template cert-manager-approver-policy jetstack/cert-manager-approver-policy \
    | yq e '. | select(.kind != "CustomResourceDefinition")'
```

If the goal is to allow quick and simple "default" installation of both CRDs and other stuff, for users who are evaluating approver-policy, why don't we put the CRDs in the `crds/` directory as recommended by Helm.
And then write upgrade instructions for production users telling them that they will have to "upgrade" the CRDs separately?

inteon:
fixed in https://github.com/cert-manager/approver-policy/pull/376

PR #322, opened 2023-12-18
Fix makefile modules self-upgrade

inteon:


wallrj (approved):
Thanks.

As I understand it, you've manually updated the `klone.yaml` file this time, 
because the self-upgrade feature had a bug.

This PR fixes the bug and updates the `klone.yaml` references to the commit hash in makefile-modules where the bug was fixed upstream.

Next time we update `makefile-modules` repo, a PR should appear in this repo automatically, looking something like:
 * https://github.com/inteon/approver-policy/pull/1

/approve
/lgtm

PR #327, opened 2023-12-19
Upgrade self-upgrade

inteon:
Bumped the `repository-base` module with improvements made to the self-upgrade script.


inteon:
/approve
self-lgtm'ing since this PR is equivalent to the self-upgrade but for the GH actions (which is not supported by the self-upgrade).
/lgtm

PR #333, opened 2024-01-02
Remove deprecated make targets

inteon:
We stopped using these targets in https://github.com/cert-manager/testing/pull/934.


wallrj (approved):
/approve
/lgtm

PR #340, opened 2024-01-12
Enable helm-tool linter and schema generator

inteon:
Enabling these two features that were added in the Helm module:
https://github.com/cert-manager/approver-policy/blob/dea7ee83b25b9778e021dae9ecaf5636aec16f90/make/00_mod.mk#L42-L43

### DIFF
--- a/deploy/charts/approver-policy/values.yaml
+++ b/deploy/charts/approver-policy/values.yaml
@@ -103,7 +103,7 @@ app:
 
       # The nodePort set on the Service used by the webhook.
       # +docs:property
-      # nodePort: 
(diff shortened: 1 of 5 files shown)


deploy/charts/approver-policy/values.yaml
```diff
@@ -93,6 +93,10 @@ app:
       # The type of Kubernetes Service used by the webhook.
       type: ClusterIP
 
+      # The nodePort set on the Service used by the webhook.
+      # +docs:property
+      # nodePort:
```
inteon, line 98: The linter found that this property was not yet documented.

PR #343, opened 2024-01-13
Add allow-all example

inteon:
I think this is quite a useful policy, eg. when migrating from cert-manager to approver-policy.
Another use case is a cluster with an allow-all namespace and another restricted namespace.

### DIFF
--- a/docs/examples/allow-all.yaml
+++ b/docs/examples/allow-all.yaml
@@ -1,0 +1,48 @@
+# This policy configures all options to allow all possible values.
+apiVersion: policy.cert-manager.io/v1alpha1
+kind: CertificateRequestPolicy
+metadata:
+  name: allow-all
+spec:
+  allowed:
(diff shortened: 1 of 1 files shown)


docs/examples/allow-all.yaml
```diff
@@ -0,0 +1,48 @@
...
+  name: allow-all
+spec:
+  allowed:
+    commonName: {value: "*"}
+    dnsNames: {values: ["*"]}
+    ipAddresses: {values: ["*"]}
+    uris: {values: ["*"]}
+    emailAddresses: {values: ["*"]}
+    isCA: true
+    usages:
```
erikgb, line 14: Wow! This is a long list! Have we ever considered adding wildcard support to this field?

erikgb (approved):
/lgtm

I think this allow-all policy example makes a lot of sense! Thanks @inteon!

PR #353, opened 2024-01-21
fix: should reconcile certificate request policies consistently

### LINKED ISSUE #341: Infinite loop causing increase in certificaterequestpolicies version
**Description**: I have encountered an issue where an infinite loop is causing an increase in the certificate request policy version. I found that the root cause of the infinite loop is related to the `setCertificateRequestPolicyCondition()` function in the certificaterequestpolicies.go file. 

First problem is here: https://github.com/cert-manager/approver-policy/blob/main/pkg/internal/controllers/certificaterequestpolicies.go#L208, empty list is created which will cause `setCertificateRequestPolicyCondition()` to always append condition in the list and try to patch object in `Reconcile()`. Furthermore, I believe if object is  patched even when it is already in a ready state, that this behavior leads to an infinite loop, as the object is modified and triggers a new event for the patched object (shouldn't we patch object only if it's not ready?): 
```
// If this update doesn't contain a state transition, we don't update
// the conditions LastTransitionTime to Now()
if existingCondition.Status == condition.Status {
	condition.LastTransitionTime = existingCondition.LastTransitionTime
}

(*conditions)[idx] = condition
return
```

I've tested fix locally and something like this should be solution: 
```
func (c *certificaterequestpolicies) setCertificateRequestPolicyCondition(
	conditions *[]policyapi.CertificateRequestPolicyCondition,
	generation int64,
	condition policyapi.CertificateRequestPolicyCondition,
) bool {
	condition.LastTransitionTime = &metav1.Time{Time: c.clock.Now()}
	condition.ObservedGeneration = generation

	for idx, existingCondition := range *conditions {
		// Skip unrelated conditions
		if existingCondition.Type != condition.Type {
			continue
		}

		// If this update doesn't contain a state transition, we don't update object
		if existingCondition.Status == condition.Status {
			return false
		}

		(*conditions)[idx] = condition
		return true
	}

	// If we've not found an existing condition of this type, we simply insert
	// the new condition into the slice.
	*conditions = append(*conditions, condition)
	return true
}
```

Then reconcile should patch object only if it's not in desired state: 
```
func (c *certificaterequestpolicies) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
	result, patch, policyStatus, resultErr := c.reconcileStatusPatch(ctx, req)
	// fmt.Println("Patching object", req.Name)
	if patch {
		// fmt.Println("Patching object", req.Name)
		crp, patch, err := ssa_client.GenerateCertificateRequestPolicyStatusPatch(req.Name, req.Namespace, policyStatus)
		if err != nil {
			err = fmt.Errorf("failed to generate CertificateRequestPolicy.Status patch: %w", err)
			return ctrl.Result{}, utilerrors.NewAggregate([]error{resultErr, err})
		}

		if err := c.client.Status().Patch(ctx, crp, patch, &client.SubResourcePatchOptions{
			PatchOptions: client.PatchOptions{
				FieldManager: "approver-policy",
				Force:        pointer.Bool(true),
			},
		}); err != nil {
			err = fmt.Errorf("failed to apply CertificateRequestPolicy.Status patch: %w", err)
			return ctrl.Result{}, utilerrors.NewAggregate([]error{resultErr, err})
		}
	}

	return result, resultErr
}
```
**How to reproduce**:
1. Start the cert-manager-approver-policy application and monitor the resource version at regular intervals (e.g., every x.y seconds).
2. Enable higher log verbosity and observe that the object is consistently being patched.

If you confirm that this is indeed a bug, I am willing to assist in fixing it and creating a pull request (PR) if you need assistance.
/kind bug

hrbasic:
```release-note
Prevent an infinite loop when patching the status of certificate request policy. 
```
Fixes #341

### DIFF
--- a/pkg/internal/controllers/certificaterequestpolicies.go
+++ b/pkg/internal/controllers/certificaterequestpolicies.go
@@ -142,7 +142,7 @@ func addCertificateRequestPolicyController(_ context.Context, opts Options) erro
 func (c *certificaterequestpolicies) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	result, patch, resultErr := c.reconcileStatusPatch(ctx, req)
 	if patch != nil {
-		crp, patch, err := ssa_client.GenerateCertificateRequestPolicyStatusPatch(req.Name, req.Namespace, patch)
+		crp, patch, err := ssa_client.GenerateCertificateRequestPolicyStatusPatch(req.Name, patch)
 		if err != nil {
 			err = fmt.Errorf("failed to generate CertificateRequestPolicy.Status patch: %w", err)
 			return ctrl.Result{}, utilerrors.NewAggregate([]error{resultErr, err})
@@ -214,6 +214,7 @@ func (c *certificaterequestpolicies) reconcileStatusPatch(ctx context.Context, r
 		c.recorder.Event(policy, corev1.EventTypeWarning, "NotReady", message)
 
 		c.setCertificateRequestPolicyCondition(
+			policy.Status.Conditions,
 			&policyPatch.Conditions,
 			policy.Generation,
 			policyapi.CertificateRequestPolicyCondition{
@@ -233,6 +234,7 @@ func (c *certificaterequestpolicies) reconcileStatusPatch(ctx context.Context, r
 	c.recorder.Event(policy, corev1.EventTypeNormal, "Ready", message)
 
 	c.setCertificateRequestPolicyCondition(
+		policy.Status.Conditions,
 		&policyPatch.Conditions,
 		policy.Generation,
 		policyapi.CertificateRequestPolicyCondition{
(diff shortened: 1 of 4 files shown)


pkg/internal/controllers/certificaterequestpolicies.go
```diff
@@ -118,6 +119,9 @@ func addCertificateRequestPolicyController(_ context.Context, opts Options) erro
 
 	return ctrl.NewControllerManagedBy(opts.Manager).
 		For(new(policyapi.CertificateRequestPolicy)).
+		// Reconcile the object only if there is a change in the spec or label.
+		// This prevents an infinite reconcile loop when the status is updated.
+		WithEventFilter(predicate.Or(predicate.GenerationChangedPredicate{}, predicate.LabelChangedPredicate{})).
```
erikgb, line 124: Why do we need events when policy labels are changed?

pkg/internal/controllers/certificaterequestpolicies.go
```diff
@@ -224,7 +226,7 @@ func (c *certificaterequestpolicies) reconcileStatusPatch(ctx context.Context, r
 			},
 		)
 
-		return result, policyPatch, nil
+		return result, &policy.Status, nil
```
erikgb, line 229: I think the logic must be more delicate than this. How do we achieve the removal of conditions that are no longer relevant? We user server-side apply (SSA) in this controller, and the idea is to set just the fields we care for. I agree we must consider existing conditions, but I still think we should prepare a status patch.

erikgb (changes requested):
Thanks for your PR, @hrbasic! I think it needs a bit more work before it is good to merge. Especially since this operator is migrated to server-side apply (SSA).

/ok-to-test

pkg/internal/controllers/certificaterequestpolicies.go
hrbasic: I wanted to cover possible future cases, but you are right, it's not need, will remove it.

pkg/internal/controllers/certificaterequestpolicies.go
hrbasic: I understand, it makes sense. I'll make changes.

pkg/internal/controllers/certificaterequestpolicies.go
```diff
@@ -256,30 +258,30 @@ func (c *certificaterequestpolicies) reconcileStatusPatch(ctx context.Context, r
 // Returns true if the condition has been updated or an existing condition has
 // been updated. Returns false otherwise.
 func (c *certificaterequestpolicies) setCertificateRequestPolicyCondition(
```
inteon, line 260: @hrbasic I think you could use this function as an example here: https://github.com/cert-manager/approver-policy/blob/eafcc19aeb7e131047378a043b37a88a8661b61a/pkg/internal/controllers/certificaterequests.go#L253-L305

pkg/internal/controllers/certificaterequestpolicies.go
hrbasic: As agreed with Tim, I've removed predicate since problem was related with updating `LastTransitionTime` and setting ns for cluster wide object: ` b.WithNamespace(namespace)`.

PR #371, opened 2024-02-06
docs(api): fenced code block to fix generated API docs

erikgb:
Adding a simple fenced block around the example in docs. It seems to be the current blocker to my WIP PR updating the API docs on our website, https://github.com/cert-manager/website/pull/1415. See https://app.netlify.com/sites/cert-manager-website/deploys/65c25dc6f909f600085e39ac for details. It seems like fenced blocks can pass more easily through the doc generator.

I'll also be ok removing the repeated example from the docs - if you think that's better. Right now my top priority is to get updated approver-policy API docs on our website.


pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -249,9 +249,10 @@ type ValidationRule struct {
 	// `cr` (map) variable to the CEL expression containing `namespace` and
 	// `name` of the `CertificateRequest` resource.
 	//
-	// Examples:
-	// - Rule for namespaced DNSNames:
-	//   {"rule": "self.endsWith(cr.namespace + '.svc.cluster.local'"}
+	// Example (rule for namespaced DNSNames):
+	// ```
+	// rule: self.endsWith(cr.namespace + '.svc.cluster.local'
+	// ```
```
SgtCoDFish, line 255: question: Is this actually valid CEL syntax? I've not really played with it much but it looks wrong, I'd expect the following:

```suggestion
	// Example (rule for namespaced DNSNames):
	// ```
	// rule: self.endsWith(cr.namespace + '.svc.cluster.local')
	// ```
```

erikgb: 🙈 Thanks, Ash!

SgtCoDFish (approved):
/lgtm
/approve

Thanks very much 😁

PR #388, opened 2024-02-29
Allow replicaCount to be set to int or string

wallrj:
This allows @erikgb to continue to use a variable string for the replicaCount field.

I added a [helm-tool type tag](https://github.com/cert-manager/helm-tool?tab=readme-ov-file#tags) to the `replicaCount` value of `number,string,null`, which has the effect of removing the type from the generated Helm JSON schema.
Probably because helm-tool doesn't recognise the type value.

As [discussed on Slack](https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1708937118648319) .

```sh
$ helm template deploy/charts/approver-policy/ | grep replicas
  replicas: 1
```

```sh
$ helm template deploy/charts/approver-policy/ --set replicaCount=2 | grep replicas
  replicas: 2
```

```sh
$ helm template deploy/charts/approver-policy/ --set replicaCount='${REPLICAS_OVERRIDE:=3}' | grep replicas
  replicas: ${REPLICAS_OVERRIDE:=3}
```

```sh
$ helm template deploy/charts/approver-policy/ --set replicaCount=null | grep replicas
  replicas:
```

### DIFF
--- a/deploy/charts/approver-policy/README.md
+++ b/deploy/charts/approver-policy/README.md
@@ -19,8 +19,8 @@ This option decides if the CRDs should be installed as part of the Helm installa
 > true
 > ```
 
-This option makes it so that the "helm.sh/resource-policy": keep annotation is added to the CRD. This will prevent Helm from uninstalling the CRD when the Helm release is uninstalled. WARNING: when the CRDs are removed, all cert-manager custom resources  
-(Certificates, Issuers, ...) will be removed too by the garbage collector.
+This option makes it so that the "helm.sh/resource-policy": keep annotation is added to the CRD. This will prevent Helm from uninstalling the CRD when the Helm release is uninstalled. WARNING: when the CRDs are removed, all cert-manager-approver-policy custom resources  
+(CertificateRequestPolicy) will be removed too by the garbage collector.
 #### **replicaCount** ~ `number,string,null`
 > Default value:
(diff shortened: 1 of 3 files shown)


erikgb:
/lgtm

Very nice, @wallrj! Thanks a lot! Are you able to cut a new alpha release so I can verify it solves my issue, or is that too much work?

deploy/charts/approver-policy/README.md
wallrj, line 1: image

wallrj:
> Very nice, @wallrj! Thanks a lot! Are you able to cut a new alpha release so I can verify it solves my issue, or is that too much work? Maybe we should be optimistic and head for a release - with an eventual bugfix if it doesn't work?

Thanks @erikgb. Yes I will do another alpha for you to test this before we release 0.13

inteon:
/approve
/lgtm
We might make this field type more strict in the future.

wallrj:
@erikgb Here's another alpha release, as promised:
 * https://github.com/cert-manager/approver-policy/releases/tag/v0.13.0-alpha.1
 * https://charts.jetstack.io/index.yaml

Please test it if you have time.

erikgb:
> Please test it if you have time.

@wallrj I have verified it works now. Thanks for fixing this issue!

Issue #207, opened 2023-03-03
Setting .Values.nameOverride makes the pod not have rights to update secret cert-manager-approver-policy-tls


smuda:
# Description
When setting helm parameter `.Values.nameOverride` to anything else than it's default value `cert-manager-approver-policy` the approver fails to generate it's tls certificate during startup.

The role allows access to one secret with a specific name which (when `.Values.nameOverride` is set to smuda) would be smuda-tls. However, in `pkg/internal/webhook/tls/tls.go` the name of the secret seems hard coded to `cert-manager-approver-policy-tls`.

To reproduce:
```shell
helm repo add jetstack https://charts.jetstack.io 
helm install cert-manager-approver jetstack/cert-manager-approver-policy --set nameOverride=smuda
```

# Expected result

That the approver pod would startup and respond happily to the readiness-probe.

# Result

The approver pod looks for and tries to update secret `cert-manager-approver-policy-tls` while the role allows `smuda-tls`. The pod is unhappy.

```
I0303 17:47:18.371313       1 webhook.go:67] webhook "msg"="running tls bootstrap process..." 
W0303 17:47:18.373066       1 reflector.go:424] pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: failed to list *v1.Secret: secrets "cert-manager-approver-policy-tls" is forbidden: User "system:serviceaccount:addon-cert-manager:smuda" cannot list resource "secrets" in API group "" in the namespace "addon-cert-manager"
E0303 17:47:18.373122       1 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: Failed to watch *v1.Secret: failed to list *v1.Secret: secrets "cert-manager-approver-policy-tls" is forbidden: User "system:serviceaccount:addon-cert-manager:smuda" cannot list resource "secrets" in API group "" in the namespace "addon-cert-manager"
E0303 17:47:19.378513       1 tls.go:130] webhook/tls "msg"="failed to generate initial serving certificate, retrying..." "error"="failed verifying CA keypair: tls: failed to find any PEM data in certificate input" "interval"="1s"
[...]

PR #390, opened 2024-03-01
Document the nameOverride  Helm chart value and add it to the JSON schema

wallrj:
We use the `nameOverride` field in the values.yaml of the Venafi enterprise plugin for approver-policy, to change the name of the Kubernetes resources deployed by the sub-chart.
But the new Helm JSON schema validation doesn't include that field, resulting in the following error:

```sh
  Release "approver-policy-enterprise" does not exist. Installing it now.
  Error: values don't meet the specifications of the schema(s) in the following chart(s):
  cert-manager-approver-policy:
  - (root): Additional property nameOverride is not allowed
...
```

`nameOverride` is a standard field that is created by `helm create`, but it is not documented in the scaffolded chart.
* https://github.com/helm/helm/blob/e007c900ce5f2233c8583565d5ba1b0433b1a213/pkg/chartutil/create.go#L433-L449

```
const defaultHelpers = `{{/*
Expand the name of the chart.
*/}}
{{- define "<CHARTNAME>.name" -}}
{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
{{- end }}


{{/*
Create a default fully qualified app name.
We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
If release name contains chart name it will be used as a full name.
*/}}
{{- define "<CHARTNAME>.fullname" -}}
{{- if .Values.fullnameOverride }}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
{{- else }}
{{- $name := default .Chart.Name .Values.nameOverride }}
{{- if contains $name .Release.Name }}
{{- .Release.Name | trunc 63 | trimSuffix "-" }}
{{- else }}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
{{- end }}
{{- end }}
{{- end }}

```

Our version is here:

https://github.com/cert-manager/approver-policy/blob/c1b51af39fdc454e9ab1ff3a11839b2d418b21a9/deploy/charts/approver-policy/templates/_helpers.tpl#L1-L6

There's another standard field called `fullNameOverride` which is added by `helm create` but which is missing from the approver-policy chart, and I don't know why.
The purpose of these fields are explained here:
 * https://stackoverflow.com/questions/63838705/what-is-the-difference-between-fullnameoverride-and-nameoverride-in-helm

Relatedly while learning more about this field I found that @smuda has reported a problem with the `nameOverride` field, but I haven't investigated that here.  
 * https://github.com/cert-manager/approver-policy/issues/207

### DIFF
--- a/deploy/charts/approver-policy/values.yaml
+++ b/deploy/charts/approver-policy/values.yaml
@@ -1,3 +1,8 @@
+# nameOverride replaces the name of the chart in the Chart.yaml file, when this
+# is used to construct Kubernetes object names.
+# +docs:property
+# nameOverride: approver-policy
+
 crds:
   # This option decides if the CRDs should be installed
   # as part of the Helm installation.
--- a/deploy/charts/approver-policy/README.md
+++ b/deploy/charts/approver-policy/README.md
@@ -6,6 +6,10 @@
 
 <!-- AUTO-GENERATED -->
 
+#### **nameOverride** ~ `string`
+
+nameOverride replaces the name of the chart in the Chart.yaml file, when this is used to construct Kubernetes object names.
+
 #### **crds.enabled** ~ `bool`
 > Default value:
 > ```yaml
--- a/deploy/charts/approver-policy/values.linter.exceptions
+++ b/deploy/charts/approver-policy/values.linter.exceptions
@@ -1,5 +1,3 @@
-value missing from values.yaml: nameOverride
-
 # Some false postives
 # See https://github.com/cert-manager/helm-tool/issues/27
 value missing from templates: tolerations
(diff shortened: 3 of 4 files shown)


deploy/charts/approver-policy/values.yaml
```diff
@@ -1,3 +1,8 @@
+# nameOverride replaces the name of the chart in the Chart.yaml file, when this
+# is used to construct Kubernetes object names.
+# +docs:property
+# nameOverride: approver-policy
```
wallrj, line 4: See 
* https://stackoverflow.com/a/63839389

https://github.com/cert-manager/approver-policy/blob/c1b51af39fdc454e9ab1ff3a11839b2d418b21a9/deploy/charts/approver-policy/templates/_helpers.tpl#L1-L6

inteon:
Thanks for the fix @wallrj
/approve
/lgtm

PR #409, opened 2024-03-25
Helm: add http_proxy, https_proxy, and no_proxy

maelvls:
The cert-manager Helm chart supports setting `HTTPS_PROXY` env variables since https://github.com/cert-manager/cert-manager/issues/310. But these Helm values are missing in approver-policy. I propose to add them.

> [!NOTE]
>
> Approver-policy doesn't do any external calls that would require an HTTP proxy (for people that need an HTTP proxy for outgoing HTTP traffic), so these new Helm parameters won't be helpful to anyone that is using the upstream version of approver-policy. These Helm parameters are useful for folks that rely on plugins that connect to the internet.

To test this, you can run:

```bash
helm template deploy/charts/venafi-enhanced-issuer \
  --set venafiEnhancedIssuer.manager.image.repository=foo \
  --set venafiEnhancedIssuer.https_proxy=http://localhost:9090
```

You will see:

```yaml
      containers:
      - name: venafi-enhanced-issuer
        env:
          - name: HTTPS_PROXY
            value: http://localhost:9090
```

[VC-31970]: https://venafi.atlassian.net/browse/VC-31970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

```bash
helm template ./deploy/charts/approver-policy/ \
  --set http_proxy=http://localhost:9090 | grep -A10 env:
```

returns:
```yaml
      - name: cert-manager-approver-policy
        env:
        - name: HTTP_PROXY
          value: http://localhost:9090
```

Internal link: [VC-31970](https://venafi.atlassian.net/browse/VC-31970)

### DIFF
--- a/deploy/charts/approver-policy/README.md
+++ b/deploy/charts/approver-policy/README.md
@@ -10,6 +10,18 @@
 
 nameOverride replaces the name of the chart in the Chart.yaml file, when this is used to construct Kubernetes object names.
 
+#### **http_proxy** ~ `string`
+
+Configures the HTTP_PROXY environment variable where a HTTP proxy is required.
+
+#### **https_proxy** ~ `string`
+
+Configures the HTTPS_PROXY environment variable where a HTTP proxy is required.
+
+#### **no_proxy** ~ `string`
+
+Configures the NO_PROXY environment variable where a HTTP proxy is required, but certain domains should be excluded.
+
 #### **crds.enabled** ~ `bool`
 > Default value:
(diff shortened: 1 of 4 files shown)


SgtCoDFish (approved):
/lgtm
/approve

I agree that this makes sense to support plugins, thanks!

PR #413, opened 2024-03-27
Fix linters

inteon:
Reduces the list of ignored linters by fixing the go code.

### DIFF
--- a/pkg/internal/cmd/options/options.go
+++ b/pkg/internal/cmd/options/options.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/client-go/rest"
 	cliflag "k8s.io/component-base/cli/flag"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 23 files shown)


pkg/internal/cmd/options/options.go
```diff
@@ -98,8 +97,10 @@ func (o *Options) Prepare(cmd *cobra.Command, approvers ...approver.Interface) *
 
 func (o *Options) Complete() error {
 	klog.InitFlags(nil)
-	log := klogr.New()
-	flag.Set("v", o.logLevel)
+	log := klog.TODO()
+	if err := flag.Set("v", o.logLevel); err != nil {
```
inteon, line 101: I verified that the `--log-level` flag still works.

SgtCoDFish (approved):
/lgtm
/approve

Cheers!

PR #433, opened 2024-04-26
Fix gosec errors and enable gosec linter

inteon:


pkg/internal/approver/manager/review.go
```diff
@@ -125,6 +125,7 @@ func (m *mngr) Review(ctx context.Context, cr *cmapi.CertificateRequest) (manage
 		)
 
 		for _, evaluator := range m.evaluators {
+			// #nosec G601 -- False positive. We expect the implementation to not keep this pointer path the call scope.
```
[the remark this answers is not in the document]
inteon: Ah, it should have been "keep this pointer past the call scope"

pkg/internal/approver/manager/review.go
```diff
@@ -125,6 +125,7 @@ func (m *mngr) Review(ctx context.Context, cr *cmapi.CertificateRequest) (manage
 		)
 
 		for _, evaluator := range m.evaluators {
+			// #nosec G601 -- False positive. The function does not keep this pointer past it's scope scope.
```
SgtCoDFish, line 128: suggestion: Couple of typos still - sorry to be a pain, I know it's just a comment

```suggestion
			// #nosec G601 -- False positive. The function does not keep this pointer past its scope.
```

inteon: It's my mistake, should have been a bit more diligent here.

PR #486, opened 2024-08-16
chore: pre-fix new nilnil linter error

erikgb:
For https://github.com/cert-manager/approver-policy/pull/485 to go through, we have to do something, since the newer version of golangci-lint adds some additional rules. Instead of just suppressing the lint errors, I suggest making the (unused) stub functions return a "not implemented" error.

/cc @inteon


inteon (commented):
/approve
/lgtm

PR #487, opened 2024-08-16
Bump sigs.k8s.io/controller-runtime to 0.19.0

erikgb:
Same problem (and solution) as described in https://github.com/cert-manager/trust-manager/pull/414. I might look more into the details here, but let's quick-fix it for now.

/cc @inteon


inteon (commented):
/approve
/lgtm

Issue #506, opened 2024-09-30
Add certificate request username/serviceaccount as an option for CEL validation


jamesglennan:
Hi Team,

I wanted to ask about getting some of the capability we get in the csi-driver-spiffe approver where it checks the service account (username field) that requested a certificate before approving the certificate request.

## Background
Currently the CEL validator only provides the name and namespace as fields to be used for validating- we want to be able to extract the serviceaccount name as well, so we can approve certificates that contain fields related to the requestor

## Why?
We are looking to take advantage of the csi-driver and issue certificates that both contain dnsNames for a given pod/service and also the URIs associated with a given ServiceAccount's SPIFFE SVID. We have a limitation with our setup- expecting a single certificate for both client and server communication. As such we want to migrate away from the csi-driver-spiffe, but we want to retain the approval policy that it offers. By having the serviceaccount name available to us, we can use CEL to validate our approvals with something like `self.endswith(cr.namespace + '/sa/' + cr.serviceaccount)` or `self.matches('spiffe://trust.domain/ns/' + cr.namespace + '/sa/' + cr.serviceaccount)`

## Solutions
I had a quick look last week in what would need to be changed- namely adding a field to the struct/protobuf and parsing the username field. Obviously I would love to hear feedback on this idea and get some input from the team about this problem.

## Technical Questions
Would we just parse the username field similar to the csi-driver-spiffe-approver and insert it into a field called `ServiceAccount`? Surely there are circumstances where the requesting username isn't in the format `system:serviceaccount:ns:name` and in those cases is it appropriate to leave the field blank. Do we provide both, have a field for username and serviceaccount, or offer some other pre-parsed version?

I look forward to discussing- I would be happy to help with the actually implementing (if the solution isn't crazy complicated) and please let me know if there's any extra details or questions about this.

SgtCoDFish:
Copying my details [from slack](https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1727676634134599)

It's important to note that csi-driver will use its own ServiceAccount to create CertificateRequest resources by default, in contrast to csi-driver-spiffe which uses the pod's SA.

csi-driver can be configured to use the pod's SA using the `app.driver.useTokenRequest=true` Helm value at install time. 

For illustration, here's a CR created by csi-driver _without_ `useTokenRequest`:

```yaml
apiVersion: cert-manager.io/v1
kind: CertificateRequest
metadata:
  creationTimestamp: "2024-09-30T08:26:02Z"
[...]

PR #514, opened 2024-10-07
Add CertificateRequest username to CEL Validator with serviceaccount functions

jamesglennan:
Closes https://github.com/cert-manager/approver-policy/issues/506

This PR is to add the username associated with a certificate request into the CEL Validator, so it is possible to approve or deny certificate requests based on who/what requested them. This is useful specifically for validating SPIFFE SVIDs where the URI field is in the format of `spiffe://trust.domain/ns/serviceAccountNamespace/sa/serviceAccountName`


SgtCoDFish:
/ok-to-test

(I don't think tests will run on a draft PR, but at least now they'll run when this is marked ready for review!)

SgtCoDFish:
I think this looks like a great start. I had a brief review - this isn't really my area of expertise but it looks how I'd expect this to look! 👍

pkg/internal/approver/validation/validator_test.go
```diff
@@ -91,3 +95,48 @@ func newCertificateRequest(namespace string) cmapi.CertificateRequest {
 	request.SetNamespace(namespace)
 	return request
 }
+
+func Test_Validator_Validate_ServiceAccount(t *testing.T) {
+	v := &validator{expression: "ServiceAccount(cr.username).isServiceAccount() && self.startsWith('spiffe://acme.com/ns/%s/sa/%s'.format([ServiceAccount(cr.username).getNamespace(),ServiceAccount(cr.username).getName()]))"}
```
erikgb, line 100:
```suggestion
	v := &validator{expression: "isServiceAccount(cr.username) && self.startsWith('spiffe://acme.com/ns/%s/sa/%s'.format([serviceAccount(cr.username).getNamespace(),serviceAccount(cr.username).getName()]))"}
```

I would suggest a slightly modified API to this new feature. Please let me know what you think and if it is possible to implement. I will await a review of the code/implementation until we are aligned on how this will will appear for the user. 😺

erikgb (changes requested):
This looks great! Thanks @jamesglennan! I've done an initial review, and I like the way this is heading!

/approve
/hold

pkg/internal/approver/validation/validator_test.go
jamesglennan: Added you changes and modified the tests- I also havent had a chance to test this in a real cluster yet

pkg/internal/approver/validation/serviceaccount.go
```diff
@@ -0,0 +1,163 @@
...
+	SAType = cel.ObjectType("cm.io.policy.pkg.internal.approver.validation.ServiceAccount")
+)
+
+// ConvertToNative implements ref.Val.ConvertToNative.
+func (sa ServiceAccount) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
+	if reflect.TypeOf(sa).AssignableTo(typeDesc) {
+		return sa, nil
+	}
+	if reflect.TypeOf("").AssignableTo(typeDesc) {
+		return sa, nil
```
erikgb, line 49: Is this correct? I would assume this should return a string. Maybe add a `String()` function to ServiceAccount type, returning the standard Kubernetes SA string: `system:serviceaccount:<namespace>:<name>`? WDYT?

pkg/internal/approver/validation/serviceaccount.go
```diff
@@ -0,0 +1,163 @@
...
+// ConvertToType implements ref.Val.ConvertToType.
+func (sa ServiceAccount) ConvertToType(typeVal ref.Type) ref.Val {
+	switch typeVal {
+	case SAType:
+		return sa
+	case types.TypeType:
+		return SAType
+	}
+	return types.NewErr("type conversion error from '%s' to '%s'", SAType, typeVal)
+}
```
erikgb, line 63: Looking at the upstream URL type, this function looks a bit different. Why?

https://github.com/kubernetes/apiserver/blob/1ea7befac323763d28173ed613a9e73ef1ef9c16/pkg/cel/url.go#L53-L61

pkg/internal/approver/validation/serviceaccount.go
```diff
@@ -0,0 +1,163 @@
...
+
+// Equal implements ref.Val.Equal.
+func (sa ServiceAccount) Equal(other ref.Val) ref.Val {
+	otherDur, ok := other.(ServiceAccount)
```
erikgb, line 67:
```suggestion
	otherSA, ok := other.(ServiceAccount)
```

Or does `otherDur` have a special meaning I don't understand? 😆

pkg/internal/approver/validation/serviceaccount.go
```diff
@@ -0,0 +1,163 @@
...
+func stringToServiceAccount(arg ref.Val) ref.Val {
+	s, ok := arg.Value().(string)
+	if !ok {
+		return types.MaybeNoSuchOverloadErr(arg)
+	}
+
+	ns, name, err := serviceaccount.SplitUsername(s)
+
+	if err != nil {
+		return ServiceAccount{
```
erikgb, line 108: Shouldn't we return an error here instead? An error here will be extremely hard to debug with the current lenient handling.

erikgb (commented):
Sorry for the delayed review here! 😊 I have added some comments and suggestions. Hoping we can get this merged soon!

pkg/internal/approver/validation/serviceaccount.go
jamesglennan: Its the same, the difference is that in the URL library `types.NewTypeValue("kubernetes.URL")` is deprecated, so i used the new function.

If I'm honest I'm not sure what the distinction between

```
	URLObject = decls.NewObjectType("kubernetes.URL")
	typeValue = types.NewTypeValue("kubernetes.URL")
	URLType   = cel.ObjectType("kubernetes.URL")
```

They all seem to return the same thing?

pkg/internal/approver/validation/serviceaccount.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"k8s.io/apiserver/pkg/authentication/serviceaccount"
+)
+
+func ServiceAccountLib() cel.EnvOption {
+	return cel.Lib(&saLib{})
+}
+
+type saLib struct{}
```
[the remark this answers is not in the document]
jamesglennan: I'm not sure what you mean by make the type as receiver? The saLib struct needs methods to compile the new CEL functions and the ServiceAccount struct is to store and convert back and forth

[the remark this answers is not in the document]
erikgb: AFAIK, it's common to keep the methods on a struct type after the struct definition in Golang. I am addressing these methods: https://github.com/cert-manager/approver-policy/blob/5ccd683693ff056880eb1388a33f2b6f13d0c688/pkg/internal/approver/validation/serviceaccount.go#L150-L160

https://go.dev/tour/methods/1

pkg/internal/approver/validation/serviceaccount.go
erikgb: Ok, thanks!

SgtCoDFish (approved):
/unhold
/lgtm
/approve

Thanks so much for this James! I agree with Erik that this looks great, and we discussed his nit in standup today and it's not a blocker.

Intent here is to release an alpha version of approver-policy for you to test and then if that works for you after a while we'll release it properly.

First step is to merge, so I'll do that now :D

erikgb (approved):

PR #525, opened 2024-11-08
WIP: test(e2e): rework to facilitate adding more tests

erikgb:


test/smoke/tests.go
```diff
@@ -37,143 +36,125 @@ import (
...
-			ObjectMeta: metav1.ObjectMeta{
-				GenerateName: "smoke-test-policy-",
-				Namespace:    namespace.Name,
-			},
-			Rules: []rbacv1.PolicyRule{
-				{APIGroups: []string{"policy.cert-manager.io"}, Resources: []string{"certificaterequestpolicies"}, ResourceNames: []string{"smoke-test-policy"}, Verbs: []string{"use"}},
-			},
-		}
-		Expect(cl.Create(ctx, &role)).NotTo(HaveOccurred())
+	Context("with allowed CommonName", func() {
```
wallrj, line 113:
```suggestion
	Context("CommonName Policy Checks", func() {
```

Because this tests both allowed and denied common names.

wallrj (approved):
Thanks @erikgb 

You've explained that this refactoring is designed to allow more smoke tests to be added.
1. I think that's a pretty good idea as long as we try to limit the number of these slow smoke tests in favour of using faster integration tests and unit-tests.
1. I think I agree that the RBAC tests probably don't prove anything right now....but it seems strange that they got added in the first place. Perhaps these predate the namespace selector feature and issuer selector features.
1. I think it *is* possible to test the RBAC "use" feature by adding impersonation config when we instantiate the controller-runtime client.
1. I ran the tests on my laptop and they passed

```
$ make test-smoke
...
Will run 2 of 2 specs
••

Ran 2 of 2 Specs in 0.394 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 2.044638932s
Test Suite Passed
```

/approve
/lgtm
/hold in case you want to do any more work on this PR before merging.

test/smoke/tests.go
erikgb: True, but the policy specifies allowed CNs. The way Ginkgo works, strings are appended into meaningful sentences (BDD) if a test fails. I think we can leave this as is, at least for now. Thanks!

erikgb:
Thanks for the review @wallrj. I think this PR can be merged as is now. I am eager to continue with the bugfix motivating me to do this.

/unhold

PR #534, opened 2024-11-24
BUGFIX: Webhook CA Secret name should match Helm templated RBAC

erikgb:
It seems like we forgot an important detail when tightening the RBAC to secrets in controller namespace. 😅 

Fixes https://github.com/cert-manager/approver-policy/issues/207


inteon:
The fix makes sense, the Secret name should match the name override value and match the RBAC.

/approve
/lgtm

PR #543, opened 2024-12-09
Add Helm chart OCI release to GH automation

inteon:
Adds logic for pushing the Helm chart to an OCI registry (and adding a non-v-prefixed tag)

You can test the released OCI Helm chart here:
`helm install approver-policy oci://quay.io/jetstack/charts/cert-manager-approver-policy --version 0.18.0-alpha.0`
`helm install approver-policy oci://quay.io/jetstack/charts/cert-manager-approver-policy --version v0.18.0-alpha.0`

See https://github.com/cert-manager/makefile-modules/pull/219 for more info.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -8,7 +8,7 @@ env:
   VERSION: ${{ github.ref_name }}
 
 jobs:
-  build_images:
+  build_and_push:
     runs-on: ubuntu-latest
 
     permissions:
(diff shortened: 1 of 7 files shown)


inteon:
> Any reason why we cannot migrate away from the now obsolete _jetstack_ name? Why not `oci://quay.io/cert-manager/charts` instead of `oci://quay.io/jetstack/charts`?

That will be part of a future change. Right now, the jetstack charts will point to jetstack OCI images.
Also, the goal is not yet for the charts to be consumed directly by the end-user, we will not advertise them.
Instead, the charts will be used as a source for the `https://charts.jetstack.io` Helm repo.

wallrj (approved):
Thanks for updating release docs and for the test release.

I agree with the strategy for eventually moving the charts to quay.io/cert-manager,
which you've described to @erikgb in the comments.

/approve
/lgtm

PR #572, opened 2025-01-29
Fix private key algorithm constraint always erroring when set

solidDoWant:
It looks like the upstream golang crypto/x509 returns a `PublicKey` instance for ed25519 instead of a `PublicKey` pointer for some reason (see [here](https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/crypto/x509/parser.go;l=282). As a result, constraint evaluation fails on all CRPs with a private key constraint being set, when the CSR being evaluated uses an ed25519 key.

This fixes the issue and adds a test to cover it.

### DIFF
--- a/pkg/internal/approver/constraints/evaluator.go
+++ b/pkg/internal/approver/constraints/evaluator.go
@@ -116,7 +116,7 @@ func decodePublicKey(pub interface{}) (cmapi.PrivateKeyAlgorithm, int, error) {
 	case *ecdsa.PublicKey:
 		return cmapi.ECDSAKeyAlgorithm, pubKey.Curve.Params().BitSize, nil
 
-	case *ed25519.PublicKey:
+	case ed25519.PublicKey:
 		return cmapi.Ed25519KeyAlgorithm, -1, nil
 
 	default:
--- a/pkg/internal/approver/constraints/evaluator_test.go
+++ b/pkg/internal/approver/constraints/evaluator_test.go
@@ -35,8 +35,9 @@ import (
 
 func Test_Evaluate(t *testing.T) {
 	var (
-		ecdsaAlg = cmapi.ECDSAKeyAlgorithm
-		rsaAlg   = cmapi.RSAKeyAlgorithm
+		ecdsaAlg   = cmapi.ECDSAKeyAlgorithm
+		ed25519Alg = cmapi.Ed25519KeyAlgorithm
+		rsaAlg     = cmapi.RSAKeyAlgorithm
 	)
 
 	tests := map[string]struct {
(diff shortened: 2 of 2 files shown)


solidDoWant:
@erikgb I can easily fix the failing linter check if you like, but it's not due to this change:

```shell
vscode ➜ .../pkg/internal/approver/constraints (main) $ git diff HEAD
diff --git a/pkg/internal/approver/constraints/evaluator_test.go b/pkg/internal/approver/constraints/evaluator_test.go
index 0550c7b..7ecc090 100644
--- a/pkg/internal/approver/constraints/evaluator_test.go
+++ b/pkg/internal/approver/constraints/evaluator_test.go
@@ -45,6 +45,13 @@ func Test_Evaluate(t *testing.T) {
                expResponse approver.EvaluationResponse
                expErr      bool
        }{
+               "copy/paste of the below test to demonstrate issue": {
+                       request: gen.CertificateRequest("", gen.SetCertificateRequestCSR(csrFrom(t, x509.ECDSA))),
+                       policy: policyapi.CertificateRequestPolicySpec{
+                               Constraints: nil,
+                       },
+                       expResponse: approver.EvaluationResponse{Result: approver.ResultNotDenied, Message: ""},
+               },
                "if no constraints defined, should return NotDenied": {
                        request: gen.CertificateRequest("", gen.SetCertificateRequestCSR(csrFrom(t, x509.ECDSA))),
                        policy: policyapi.CertificateRequestPolicySpec{
vscode ➜ .../pkg/internal/approver/constraints (main) $ golangci-lint run --go 1.23.5 -c /workspaces/approver-policy/.golangci.yaml
evaluator_test.go:175:66: `csrFrom` - `mods` always receives `nil` (unparam)
func csrFrom(t *testing.T, keyAlgorithm x509.PublicKeyAlgorithm, mods ...gen.CSRModifier) []byte {
                                                                 ^
vscode ➜ .../pkg/internal/approver/constraints (main) $
```

This same check fails on master if I add another test identical to an existing check.

How would you like me to proceed?

erikgb:
> This same check fails on master if I add another test identical to an existing check.
> 
> How would you like me to proceed?

It seems like the helper is copied from another test: https://github.com/cert-manager/approver-policy/blob/1ab7c9646f71472b8787671ec1dede51ca54c9ff/pkg/internal/approver/allowed/evaluator_test.go#L479-L486. Ideally we should use the same helper, but that seems like a bigger refactoring. I suggest that you just make the linter happy by removing the unused variadic argument from the helper.

solidDoWant:
Thanks! I did just that. Looks like the linter is passing now.

erikgb:
> It looks like the upstream golang crypto/x509 returns a `PublicKey` instance for ed25519 instead of a `PublicKey` pointer for some reason (see [here](https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/crypto/x509/parser.go;l=282). As a result, constraint evaluation fails on all CRPs with a private key constraint being set, when the CSR being evaluated uses an ed25519 key.

This is strange. I wonder why? 🤔 Do you know, @SgtCoDFish? Thanks for fixing this, @solidDoWant! 🚀 

/lgtm
/approve

PR #599, opened 2025-03-14
Set container port names on Helm chart

solidDoWant:
Setting the container port names makes writing network policies a little easier. The set names match the port names that cert-manager uses.

### DIFF
--- a/deploy/charts/approver-policy/templates/deployment.yaml
+++ b/deploy/charts/approver-policy/templates/deployment.yaml
@@ -38,8 +38,12 @@ spec:
         image: "{{ template "image" (tuple .Values.image $.Chart.AppVersion) }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         ports:
-        - containerPort: {{ .Values.app.webhook.port }}
-        - containerPort: {{ .Values.app.metrics.port }}
+        - name: https
+          containerPort: {{ .Values.app.webhook.port }}
+        - name: http-metrics
+          containerPort: {{ .Values.app.metrics.port }}
+        - name: healthcheck
+          containerPort: {{ .Values.app.readinessProbe.port }}
         readinessProbe:
           httpGet:
             port: {{ .Values.app.readinessProbe.port }}
--- a/deploy/charts/approver-policy/templates/deployment.yaml
+++ b/deploy/charts/approver-policy/templates/deployment.yaml
@@ -38,9 +38,9 @@ spec:
         image: "{{ template "image" (tuple .Values.image $.Chart.AppVersion) }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         ports:
-        - name: https
+        - name: webhook
           containerPort: {{ .Values.app.webhook.port }}
-        - name: http-metrics
+        - name: metrics
           containerPort: {{ .Values.app.metrics.port }}
         - name: healthcheck
           containerPort: {{ .Values.app.readinessProbe.port }}


deploy/charts/approver-policy/templates/deployment.yaml
```diff
@@ -38,8 +38,12 @@ spec:
...
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         ports:
-        - containerPort: {{ .Values.app.webhook.port }}
-        - containerPort: {{ .Values.app.metrics.port }}
+        - name: https
+          containerPort: {{ .Values.app.webhook.port }}
+        - name: http-metrics
+          containerPort: {{ .Values.app.metrics.port }}
+        - name: healthcheck
+          containerPort: {{ .Values.app.readinessProbe.port }}
```
erikgb, line 46:
```suggestion
```

Why do we need to add a named port for the health checks? I don't see the benefit.

deploy/charts/approver-policy/templates/deployment.yaml
```diff
@@ -38,8 +38,12 @@ spec:
         image: "{{ template "image" (tuple .Values.image $.Chart.AppVersion) }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         ports:
-        - containerPort: {{ .Values.app.webhook.port }}
-        - containerPort: {{ .Values.app.metrics.port }}
+        - name: https
+          containerPort: {{ .Values.app.webhook.port }}
+        - name: http-metrics
```
erikgb, line 43:
```suggestion
        - name: metrics
```

deploy/charts/approver-policy/templates/deployment.yaml
```diff
@@ -38,8 +38,12 @@ spec:
         image: "{{ template "image" (tuple .Values.image $.Chart.AppVersion) }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         ports:
-        - containerPort: {{ .Values.app.webhook.port }}
-        - containerPort: {{ .Values.app.metrics.port }}
+        - name: https
```
erikgb, line 41:
```suggestion
        - name: webhook
```

erikgb (commented):
Thanks for looking into this!

I generally prefer functional names on ports, and while we generally advice contribtors to look to cert-manager when adding missing features to side projects, I think the port names in cert-manager are a bit "historical". Some users might request enabling TLS on the metrics endpoint down the line.

If introducing named ports, I think we should benefit from the naming in other resources like the `Service`. WDYT?

deploy/charts/approver-policy/templates/deployment.yaml
solidDoWant: > Why do we need to add a named port for the health checks? I don't see the benefit.

This is so that network policies for health checks can target the functional name, rather than the port number. [Here is the netpol that I've deployed for approver-policy](https://github.com/solidDoWant/infra-mk3/blob/master/cluster/gitops/certificates/approver-policy/app/netpol.yaml#L29). Note that while this is a cilium-specific resource, the same concept applies to the general k8s networkpolicy resource as well. Whether or not probes from the host/kubelet are allowed by default is CNI-dependent.

solidDoWant:
> I generally prefer functional names on ports, and while we generally advice contribtors to look to cert-manager when adding missing features to side projects, I think the port names in cert-manager are a bit "historical". Some users might request enabling TLS on the metrics endpoint down the line.

Sure, though this would make writing more broad network policies a little more difficult. That is to say, you couldn't have a single netpol for a "certificates" namespace that would match all ports named "http-metrics" (from the cert-manager chart), and have the policy work with the approver-policy chart. However, this probably isn't a terribly common use-case.

I think your preference of functional names over historic names is the right approach. If somebody else wants historic names, another PR could be created that makes the port name configurable.

> If introducing named ports, I think we should benefit from the naming in other resources like the Service. WDYT?

Absolutely - I'll add matching names to the service.

deploy/charts/approver-policy/templates/deployment.yaml
erikgb: I see, in all the clusters I work on it must be enabled by default. 😉 Then it makes sense! But could we please use functional port names and not mix in protocols etc.?

erikgb:
> If somebody else wants historic names, another PR could be created that makes the port name configurable.

Please don't! 😆 We should try keeping things as simple as possible. And configurable port names will never make sense - at least not to me.

deploy/charts/approver-policy/templates/deployment.yaml
solidDoWant: Yea from what I've gathered, most CNIs have specific logic to allow for this. However they do see the traffic:

image

And under certain conditions require explicit policies the probe port(s). [Here's an example with](https://github.com/istio/istio/issues/49277) Cilium + Istio in eBPF mode where probes fail because Cilium expects them to come from the host, but they instead come from a link-local address.

solidDoWant:
> Please don't! 😆 We should try keeping things as simple as possible. And configurable port names will never make sense - at least not to me.

Sorry - that was a poorly-phrased way to say, if somebody else wants this, they can raise a separate issue/PR and propose the change independently of this work. In other words, I won't be changing this PR to make port names configurable (I agree with keeping this simple).

hawksight:
Thanks for submitting the improvement, it looks good to me as well. However as Prow has indicated, will need someone else to review it too

SgtCoDFish (approved):
/lgtm
/approve

Thanks everyone - looks great!

PR #603, opened 2025-03-24
Fix Ginkgo commands

erikgb:
... failing with newer versions of Ginkgo, ref. CI failing in https://github.com/cert-manager/approver-policy/pull/602.

Ginkgo CLI is now validating commands more strictly than it used to, and I think our current commands broke with https://github.com/onsi/ginkgo/releases/tag/v2.23.1.


hawksight (approved):
/lgtm

As it only seems to move the existing flags around and remove `=` this seems fine to me.

PR #606, opened 2025-04-02
Add Helm option to create RBAC allowing approval for all issuers

ali-hamza-noor:
Add Helm option to create RBAC allowing approval for all issuers


ali-hamza-noor:
Opened https://github.com/cert-manager/approver-policy/pull/607 with DCO

PR #607, opened 2025-04-02
Add Helm option to create RBAC allowing approval for all issuers

ali-hamza-noor:
Add Helm option to create RBAC allowing approval for all issuers

### DIFF
--- a/deploy/charts/approver-policy/templates/allow-all-clusterrole.yaml
+++ b/deploy/charts/approver-policy/templates/allow-all-clusterrole.yaml
@@ -1,0 +1,15 @@
+{{- if .Values.allowAllIssuerRbac -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    {{- include "cert-manager-approver-policy.labels" . | nindent 4 }}
+  name: {{ include "cert-manager-approver-policy.name" . }}-allow-all
+rules:
+  - apiGroups:
+      - cert-manager.io
+    resources:
+      - "signers"
+    verbs:
+      - "approve"
+{{- end -}}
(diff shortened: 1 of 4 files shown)


erikgb:
@ali-hamza-noor First, I would like to thank you for joining our ContribFest at KubeCon EU and working on this issue! I wanted to review it now, and when checking the existing RBAC, I think this feature is already available, ref. https://github.com/cert-manager/approver-policy/blob/d293f066bdcb5b0e08528705d39c447a05e4b705/deploy/charts/approver-policy/templates/clusterrole.yaml#L24-L32

It would be nice if you and maybe also @wallrj or @SgtCoDFish could confirm this so that we can close the issue. Seem to have been added in https://github.com/cert-manager/approver-policy/pull/416.

ali-hamza-noor:
@erikgb Thanks for having a look! it does seem like that it was already implemented. I will wait for @wallrj to confirm so we can close the issue

maelvls:
Seems like you are right, Erik. I didn't pay attention during the ContribFest: the issue has already been fixed. Ash already fixed the issue. There is no `resourceNames` by default.

For example:

```bash
helm template oci://quay.io/jetstack/charts/cert-manager-approver-policy --show-only templates/clusterrole.yaml
```

shows:

```yaml
kind: ClusterRole
rules:
- apiGroups: ["cert-manager.io"]
  resources: ["signers"]
  verbs: ["approve"]
```

The user can re-enable this extra (superfrlous, if you ask me) RBAC by using `--set app. approveSignerNames="{something}"`.

@ali-hamza-noor My bad, we should have looked into the existing ClusterRole 😅

erikgb:
This option seems already available after https://github.com/cert-manager/approver-policy/pull/416. Closing.

PR #623, opened 2025-04-24
Bump Go to 1.24 fixing new lint errors

erikgb:
In https://github.com/cert-manager/approver-policy/pull/622, we are forced to bump Go to 1.24, but that introduces new linter errors from the [`usetesting` linter](https://golangci-lint.run/usage/linters/#usetesting). Since this requires many changes, I think it's best to do it in a separate PR that is reviewed (by someone other than the commit author).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/cert-manager/approver-policy
 
-go 1.23.0
+go 1.24.0
 
 require (
 	github.com/cert-manager/cert-manager v1.17.1
--- a/pkg/internal/approver/manager/predicate/predicate_test.go
+++ b/pkg/internal/approver/manager/predicate/predicate_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package predicate
 
 import (
-	"context"
 	"path"
 	"testing"
 
@@ -38,12 +37,7 @@ import (
 )
 
 func Test_RBACBound(t *testing.T) {
-	ctx, cancel := context.WithCancel(context.TODO())
-	t.Cleanup(func() {
-		cancel()
-	})
-
-	env := testenv.RunControlPlane(t, ctx,
+	env := testenv.RunControlPlane(t, t.Context(),
 		testenv.GetenvOrFail(t, "CERT_MANAGER_CRDS"),
 		path.Join("..", "..", "..", "..", "..", "deploy", "crds"),
 	)
@@ -53,7 +47,7 @@ func Test_RBACBound(t *testing.T) {
 		requestNamespace = "test-namespace"
 	)
 
-	if err := env.AdminClient.Create(context.TODO(),
+	if err := env.AdminClient.Create(t.Context(),
 		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: requestNamespace}},
 	); err != nil {
 		t.Fatal(err)
@@ -379,11 +373,12 @@ func Test_RBACBound(t *testing.T) {
 		},
 	}
 
+	ctx := t.Context()
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
 			t.Cleanup(func() {
 				for _, obj := range test.apiObjects {
-					if err := env.AdminClient.Delete(context.TODO(), obj); err != nil {
+					if err := env.AdminClient.Delete(ctx, obj); err != nil {
 						// Don't Fatal here as a ditch effort to at least try to clean-up
 						// everything.
 						t.Errorf("failed to deleted existing object: %s", err)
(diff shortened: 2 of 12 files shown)


erikgb:
@inteon @wallrj I am struggling to fix the remaining lint errors for obtaining a `Context` inside the `Cleanup` functions. Any suggestions? Is suppression our only option?

pkg/internal/approver/manager/review_test.go
```diff
@@ -217,7 +212,8 @@ func Test_Review(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			t.Cleanup(func() {
 				for _, obj := range test.policies {
-					if err := env.AdminClient.Delete(context.TODO(), &obj); /* #nosec G601 -- Func drops pointer at end of call. */ err != nil {
+					ctx := context.TODO() //nolint:usetesting
```
erikgb, line 215: I'm not sure how to fix this without suppression - the `t.Context` is already done when entering `t.Cleanup` functions.

pkg/internal/approver/manager/predicate/predicate_test.go
```diff
@@ -383,7 +378,8 @@ func Test_RBACBound(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			t.Cleanup(func() {
 				for _, obj := range test.apiObjects {
-					if err := env.AdminClient.Delete(context.TODO(), obj); err != nil {
+					ctx := context.TODO() //nolint:usetesting
```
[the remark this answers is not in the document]
wallrj: I wondered if you could use a context from the parent test, which might only be cancelled after all the child tests have finished and cleaned up?

Or else use `context.WithoutCancel` ?

The tests seem to pass with this diff....but it may be subtly broken...not sure.

```diff
$ git diff
diff --git i/pkg/internal/approver/manager/predicate/predicate_test.go w/pkg/internal/approver/manager/predicate/predicate_test.go
index 9e016e6..cdc3642 100644
--- i/pkg/internal/approver/manager/predicate/predicate_test.go
+++ w/pkg/internal/approver/manager/predicate/predicate_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package predicate

 import (
-       "context"
        "path"
        "testing"

@@ -374,11 +373,11 @@ func Test_RBACBound(t *testing.T) {
                },
        }

+       ctx := t.Context()
        for name, test := range tests {
                t.Run(name, func(t *testing.T) {
                        t.Cleanup(func() {
                                for _, obj := range test.apiObjects {
-                                       ctx := context.TODO() //nolint:usetesting
                                        if err := env.AdminClient.Delete(ctx, obj); err != nil {
                                                // Don't Fatal here as a ditch effort to at least try to clean-up
                                                // everything.
@@ -388,7 +387,7 @@ func Test_RBACBound(t *testing.T) {
                        })

                        for _, obj := range test.apiObjects {
-                               if err := env.AdminClient.Create(t.Context(), obj); err != nil {
+                               if err := env.AdminClient.Create(ctx, obj); err != nil {
                                        t.Fatalf("failed to create new object: %s", err)
                                }
                        }
@@ -404,7 +403,7 @@ func Test_RBACBound(t *testing.T) {
                                        },
                                },
                        }
-                       policies, err := RBACBound(env.AdminClient)(t.Context(), req, test.policies)
+                       policies, err := RBACBound(env.AdminClient)(ctx, req, test.policies)
                        assert.NoError(t, err)
                        assert.Equal(t, test.expPolicies, policies)
                })
```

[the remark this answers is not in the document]
inteon: @erikgb I agree with @wallrj, you could eg. use `context.WithTimeout(context.WithoutCancel(t.Context()), 20*time.Second)`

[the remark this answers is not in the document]
erikgb: Thanks! TIL! Should be ready for another review now!

wallrj (approved):
Thanks @erikgb 

I see that the project is already being built with the latest version of Go 1.24,

https://github.com/cert-manager/approver-policy/blob/5c203a9346d02557cf30eaabc94761e766b47102/make/_shared/tools/00_mod.mk#L163-L164

So I think this makes sense and it's nice to get rid of all those context.TODO's

/approve
/lgtm

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cert-manager/approver-policy
 
-go 1.23.0
+go 1.24.0
```
wallrj, line 3: /hold Won't dependabot still add that toolchain line in #622, unless we set this to 1.24.2?

erikgb: No, I think not. But there is only one way to find out. 😉 I have reported toolchain issues to Dependabot multiple times, but the bugs keep coming back.

erikgb: The latest weeks, Dependabot seems to only add toolchain if it needs to bump the Go version. And I don't think 1.24.2 is required by our (updated) dependencies. It's just the version we use.

erikgb: Can I unhold to see what happens?

PR #628, opened 2025-04-27
Grant cert-manager RBAC to use all policies by default

erikgb:
This PR should make it a bit easier to use approver-policy with cert-manager. By default, it will now grant RBAC permissions to use all CertificatePolicies.

Close https://github.com/cert-manager/approver-policy/issues/216

/cc @hawksight @inteon @SgtCoDFish


inteon:
I think this might be something for a v2 of this component.
Also, I would rather just drop the RBAC logic.

erikgb:
> I think this might be something for a v2 of this component. Also, I would rather just drop the RBAC logic.

Could you elaborate? Especially on the last part. "RBAC logic"?

inteon:
> > I think this might be something for a v2 of this component. Also, I would rather just drop the RBAC logic.
> 
> Could you elaborate? Especially on the last part. "RBAC logic"?

Instead of using RBAC to link CertificateRequests with CertificateRequestPolicies, we can use the `.spec.selector` instead.

erikgb:
> > > I think this might be something for a v2 of this component. Also, I would rather just drop the RBAC logic.
> > 
> > 
> > Could you elaborate? Especially on the last part. "RBAC logic"?
> 
> Instead of using RBAC to link CertificateRequests with CertificateRequestPolicies, we can use the `.spec.selector` instead.

But why did you create the referenced issue? The selector is already in use, but that connects the CertificateRequestPolicy to issuers. Not to **who** created the CertificateRequest. AFAIK, RBAC isn't able to use selectors.

erikgb:
@github-copilot review

PR #712, opened 2025-09-14
Add new certificaterequests_approval metric

erikgb:
The golangci-lint exclusion configuration triggered this change; instead of just adding the lint suppressions, I attempted to determine how these metrics could be made more useful for users.

Since the existing metrics don't follow best practices for naming metrics, I propose to deprecate them with a replacement.

For the replacement, I think it's more natural to have a single gauge metric with a label indicating if the CR is approved, denied, or unmatched.

I have also removed the explicit use of the c/r Cache, and instead use the c/r cached client, as this simplifies the code to a great deal. If this change is accepted, we should probably refactor the test code a bit more, but I didn't want to change the tests too much in this PR - to make it easier reviewable.

### DIFF
--- a/pkg/internal/metrics/metrics.go
+++ b/pkg/internal/metrics/metrics.go
@@ -18,17 +18,38 @@ package metrics
 
 import (
 	"context"
-	"time"
 
 	cmapi "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	cmmeta "github.com/cert-manager/cert-manager/pkg/apis/meta/v1"
 	"github.com/go-logr/logr"
 	"github.com/prometheus/client_golang/prometheus"
-	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
 var (
+	// certificateRequestsApproval reports the number of CertificateRequest resources
+	// with approver-policy conditions set.
+	//
+	// It is a GaugeVec with the following labels:
+	//   - "namespace": the namespace of the CertificateRequest.
+	//   - "status": the approval status of the CertificateRequest.
+	//     Possible values are:
+	//       * "approved"    — the CertificateRequest has been approved.
+	//       * "denied"      — the CertificateRequest has been denied.
+	//       * "unmatched"   — the CertificateRequest not matched by any approvers.
+	//
+	// Exposed as the metric `certmanager_approverpolicy_certificaterequests_approval`.
+	certificateRequestsApproval = prometheus.NewDesc(
+		"certmanager_approverpolicy_certificaterequests_approval",
+		"The approval status of CertificateRequests. Possible values for the 'status' label: 'approved', 'denied', 'unmatched'.",
+		[]string{
+			"namespace",
+			"status",
+		},
+		nil,
+	)
+
 	// approvedCount counts the number of CertificateRequest currently approved
 	// or denied by looking at the Approved condition. For context, the Approved
 	// condition looks like this:
@@ -43,7 +64,7 @@ var (
 	// get removed over time e.g. with revisionHistoryLimit.
 	approvedCount = prometheus.NewDesc(
 		"approverpolicy_certificaterequest_approved_count",
-		"Number of CertificateRequests that have been approved (Approved=True).",
+		"DEPRECATED: use certmanager_approverpolicy_certificaterequests_approval instead. Number of CertificateRequests that have been approved (Approved=True).",
 		[]string{
 			"namespace",
 		},
@@ -60,7 +81,7 @@ var (
 	//     Invalid value: []string{"forbidden-domain-41.com"}: *.example.com, *.ca-wont-accept.org]'
 	deniedCount = prometheus.NewDesc(
 		"approverpolicy_certificaterequest_denied_count",
-		"Number of CertificateRequests that have been denied (Denied=True).",
+		"DEPRECATED: use certmanager_approverpolicy_certificaterequests_approval instead. Number of CertificateRequests that have been denied (Denied=True).",
 		[]string{
 			"namespace",
 		},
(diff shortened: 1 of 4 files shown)


pkg/internal/metrics/metrics.go
```diff
@@ -159,6 +205,7 @@ func collectCRsApproved(ctx context.Context, log logr.Logger, c cache.Cache, ch
 	}
 
 	for _, key := range labels {
+		//nolint:promlinter // This metric is deprecated and will eventually be removed.
```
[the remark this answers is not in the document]
erikgb: This makes the suppression non-functional.

pkg/internal/metrics/metrics.go
```diff
@@ -249,14 +295,12 @@ func collectCRsUnmatched(logger logr.Logger, c cache.Cache, ch chan<- prometheus
...
+func isStatusConditionTrue(conditions []cmapi.CertificateRequestCondition, conditionType cmapi.CertificateRequestConditionType) bool {
+	for _, condition := range conditions {
+		if condition.Type == conditionType {
+			return condition.Status == cmmeta.ConditionTrue
 		}
 	}
-
-	return cmmeta.ConditionUnknown
+	return false
 }
```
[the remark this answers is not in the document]
erikgb: This is kinda obvious.

maelvls:
> The golangci-lint exclusion configuration triggered this change; instead of just adding the lint suppressions, I attempted to determine how these metrics could be made more useful for users. Since the existing metrics don't follow best practices for naming metrics, I propose to deprecate them with a replacement.

Thanks for digging into that; I was the one who added the two metrics, admittedly without knowing what the best practices were, so I want to learn. 😅

For anyone looking at this, the linter warning looks like this:

> non-histogram and non-summary metrics **should not have "_count" suffix**

As Erik correctly points outs, the three metrics (two of which I added in #289) have the GAUGE type but have the suffix `_count`:

```
approverpolicy_certificaterequest_approved_count (GAUGE set to 0 or 1)
approverpolicy_certificaterequest_denied_count (GAUGE set to 0 or 1)
approverpolicy_certificaterequest_unmatched_count (GAUGE set to 0 or 1)
```

> For the replacement, I think it's more natural to have a single gauge metric with a label indicating if the CR is approved, denied, or unmatched.

Agreed. I guess I should have done in #289 what you did in your PR: it doesn't make sense to have three counters for mutually exclusive states. It makes it trickier to create dashboards.

I agree with this PR. That said, I'm a little unsure about "this metric will eventually be removed"; unless we do have a plan to remove it with at least a link to an issue explaining what the timeline is, I'd rather not say that it will be removed

Other than this, all good!

pkg/internal/metrics/metrics.go
```diff
@@ -18,17 +18,38 @@ package metrics
...
+	//
+	// It is a GaugeVec with the following labels:
+	//   - "namespace": the namespace of the CertificateRequest.
+	//   - "status": the approval status of the CertificateRequest.
+	//     Possible values are:
+	//       * "approved"    — the CertificateRequest has been approved.
+	//       * "denied"      — the CertificateRequest has been denied.
+	//       * "unmatched"   — the CertificateRequest not matched by any approvers.
+	//
+	// Exposed as the metric `certmanager_approverpolicy_certificaterequests_approval`.
```
maelvls, line 42: At first, I thought "shouldn't it use the `_total` suffix" but I realized that `_total` is meant for monotonically increasing counters. All good.

maelvls: Just to compare, here is what kube-state-metrics's pod status looks like:


```
# HELP kube_pod_status_ready [STABLE] Describes whether the pod is ready to serve requests.
# TYPE kube_pod_status_ready gauge
kube_pod_status_ready{namespace="venafi",pod="venafi-kubernetes-agent-56b8854ccd-gxpg2",uid="698e56d1-b503-4c2a-a32b-a543f3ed429f",condition="true"} 1
kube_pod_status_ready{namespace="venafi",pod="venafi-kubernetes-agent-56b8854ccd-gxpg2",uid="698e56d1-b503-4c2a-a32b-a543f3ed429f",condition="false"} 0
kube_pod_status_ready{namespace="venafi",pod="venafi-kubernetes-agent-56b8854ccd-gxpg2",uid="698e56d1-b503-4c2a-a32b-a543f3ed429f",condition="unknown"} 0
```

The gauge reflects the contents of the condition `Ready`.

In that case, shouldn't the name be smth like:

```diff
-certmanager_approverpolicy_certificaterequests_approval
+certmanager_approverpolicy_certificaterequests_status_approved
```

Although it's not a single condition, but rather two conditions (Denied=True and Approved=True)... so I guess we can't really use kube-state-metrics' convention.

pkg/internal/metrics/metrics_test.go
```diff
@@ -64,7 +148,7 @@ func Test_Metrics(t *testing.T) {
 			},
 		})
 		const expected = `
-            # HELP approverpolicy_certificaterequest_approved_count Number of CertificateRequests that have been approved (Approved=True).
+            # HELP approverpolicy_certificaterequest_approved_count DEPRECATED: use certmanager_approverpolicy_certificaterequests_approval instead. Number of CertificateRequests that have been approved (Approved=True).
```
maelvls, line 151:
```suggestion
            # HELP approverpolicy_certificaterequest_approved_count DEPRECATED: use certmanager_approverpolicy_certificaterequests_approval instead, see: https://github.com/cert-manager/approver-policy/issues/713. Number of CertificateRequests that have been approved (Approved=True).
```

pkg/internal/metrics/metrics.go
```diff
@@ -73,7 +94,7 @@ var (
 	// condition.
 	unmatchedCount = prometheus.NewDesc(
 		"approverpolicy_certificaterequest_unmatched_count",
-		"Number of CertificateRequests not matched to any policy, i.e., that don't have an Approved or Denied condition set yet.",
+		"DEPRECATED: use certmanager_approverpolicy_certificaterequests_approval instead. Number of CertificateRequests not matched to any policy, i.e., that don't have an Approved or Denied condition set yet.",
```
maelvls, line 97:
```suggestion
		"DEPRECATED: use certmanager_approverpolicy_certificaterequests_approval instead, see: https://github.com/cert-manager/approver-policy/issues/713. Number of CertificateRequests not matched to any policy, i.e., that don't have an Approved or Denied condition set yet.",
```

maelvls:
I've added a comment on adding the link to the metric's HELP too, but feel free to skip that if you don't feel it is necessary.

/lgtm
/hold if you want to make that change

erikgb:
> I've added a comment on adding the link to the metric's HELP too, but feel free to skip that if you don't feel it is necessary.

How is it useful to add the link to the HELP text? Probably something I don't understand.

maelvls:
Nevermind, folks can just look up on GitHub to know why this has been deprecated, adding the link to the HELP isn't useful.

/lgtm

Issue #797, opened 2026-01-30
Add certificate request groups to the `cr` map for CEL validation


jonasbadstuebner:
Hey there,

I saw that https://github.com/cert-manager/approver-policy/pull/514 added the username for use in `.validations`, but the `groups` of the `CertificateRequest` are still not included.

## What do I wish for?
I would like to have access to the [`groups` of `UserInfo`](https://cert-manager.io/docs/usage/certificaterequest/#userinfo) when using `CertificateRequestPolicy.spec.allowed.**.validations[index]`.

## Why?
We are evaluating ways for our colleagues to issue TLS certificates for themselves using `kubectl cert-manager create` to access products running in our clusters.
Everyone is authenticating against our clusters with `kubectl` using OIDC, so we have the groups from our centralized identity provider put on the `CertificateRequest`s. They reflect certain access levels and what team a colleague is part of.
Having access to this list of groups when verifying if a colleague is allowed to ask for a certain certificate to be issued for them would remove an extra moving part for us to verify the match in a different way outside of cert-manager and reduce the risk of getting the implementation wrong there.

## Proposed solution
I think adding the groups as an array of strings to the protobuf and giving it into the validation rule would do the trick.

jonasbadstuebner:
Hey @erikgb,
thank you for your swift reply! I opened a PR to tackle this, please let me know if something is missing.

PR #799, opened 2026-01-30
feat(approver,validation): expose groups on cr map for validations

jonasbadstuebner:
Closes https://github.com/cert-manager/approver-policy/issues/797

This PR takes the groups from the `CertificateRequest` and adds them to the `cr` map so they can be used for custom CEL validations.


pkg/internal/approver/validation/validator_test.go
```diff
@@ -140,3 +141,46 @@ func newCertificateRequestWithUsername(username string) cmapi.CertificateRequest
...
+
+	type args struct {
+		val string
+		cr  cmapi.CertificateRequest
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    bool
+		wantErr bool
```
erikgb: Nit: This field is currently not in use. Maybe remove it?

PR #855, opened 2026-03-30
fix: use %w verb for proper error wrapping

archy-rock3t-cloud:
Use `%w` instead of `%v`/`%s` in `fmt.Errorf` calls to enable `errors.Is` and `errors.As` for error inspection.


erikgb (commented):
Thanks!

/lgtm
/approve

PR #868, opened 2026-04-02
feat(annotations): Add annotation-based policy enforcement

bmsiegel:
Add a new 'annotations' field to CertificateRequestPolicyAllowed that enables administrators to define policies over CertificateRequest annotations.

The field reuses the existing CertificateRequestPolicyAllowedStringSlice type, so wildcards, required, and CEL validations work out of the box.

This check only validates that the annotations present in the policy match the annotations on the request (ie the request can have more tags than just those in this policy). If the annotations on the request are different than the annotations in the policy, the request is allowed. This is to protect administrators from having to put every annotation on their certificate request in the policy.


inteon (commented):
I think this might result in an exploitable race condition: annotations are mutable and might get changed between approval and issuance.

PR #908, opened 2026-05-21
BUGFIX: Make ValidatingWebhookConfiguration precise

### LINKED ISSUE #907: Cert-manager-approver-policy ValidatingWebhookConfiguration uses wildcard resources causing self-referential deadlock
**Summary**
The cert-manager-approver-policy ValidatingWebhookConfiguration has resources: ['*/*'] which intercepts subresource operations (including /status patches), causing the approver-policy controller to deadlock against its own webhook during reconciliation.

**Impact**
cert-manager-approver-policy pod enters CrashLoopBackOff (54+ restarts in 21h)
Any Helm chart deploying CertificateRequestPolicy resources fails with:

Error: failed to create resource: Internal error occurred: failed calling webhook "policy.cert-manager.io": failed to call webhook: Post "https://cert-manager-approver-policy.nm-cal-certificatemanagement.svc:443/validate-policy-cert-manager-io-v1alpha1-certificaterequestpolicy?timeout=5s": context deadline exceeded
Downstream deployments (e.g. nm-namespaces-production) are blocked
Root Cause
The webhook rule in ValidatingWebhookConfiguration/cert-manager-approver-policy:

rules:- apiGroups:  - policy.cert-manager.io  apiVersions:  - '*'  operations:  - CREATE  - UPDATE  resources:  - '*/*'       # <-- catches subresources including /status  scope: '*'
The wildcard */* matches certificaterequestpolicies/status. When the approver-policy reconciler patches the status of CertificateRequestPolicy objects, the API server invokes the webhook back on the same pod. During startup (TLS cert not yet generated) or under load, this self-call times out, causing:

Reconciler errors (context deadline exceeded)
Pod unable to renew leader lease → exits with "leader election lost"
CrashLoopBackOff cycle

**Temperory Fix**
Patch the webhook to target only the main resource (not subresources):

kubectl patch validatingwebhookconfiguration cert-manager-approver-policy \  --type='json' \  -p='[{"op": "replace", "path": "/webhooks/0/rules/0/resources", "value": ["certificaterequestpolicies"]}]'

erikgb:
Our ValidatingWebhookConfiguration has an incorrect (too generic) configuration, which may result in startup issues. We should probably also select the API version more precisely, but for now, let's fix the main issue of resource status update requests being intercepted by the webhook.

Fixes https://github.com/cert-manager/approver-policy/issues/907

### DIFF
--- a/deploy/charts/approver-policy/templates/webhook.yaml
+++ b/deploy/charts/approver-policy/templates/webhook.yaml
@@ -40,7 +40,7 @@ webhooks:
           - CREATE
           - UPDATE
         resources:
-          - "*/*"
+          - certificaterequestpolicies
     admissionReviewVersions: ["v1", "v1beta1"]
     timeoutSeconds: {{ .Values.app.webhook.timeoutSeconds }}
     failurePolicy: Fail


wallrj-cyberark:
@erikgb This change has been included in [v0.26.0](https://github.com/cert-manager/approver-policy/releases/tag/v0.26.0), which is now published.

Installation instructions are at https://cert-manager.io/docs/policy/approval/approver-policy/installation/

If you are able to install the new release and verify the fix in your environment, that would be much appreciated. Thank you for the contribution.

PR #915, opened 2026-06-01
Fix SAN-extension policy bypass

mladen-rusev-cyberark:
## TL;DR
A low-privileged tenant (`create certificaterequests` + RBAC `use` on a bound
`CertificateRequestPolicy`) could obtain a CA-signed certificate carrying SAN
identities the approver-policy gate never evaluated — e.g. a Microsoft UPN
`otherName` or a `directoryName` — and impersonate a privileged user to AD CS /
EAP-TLS / custom mTLS relying parties.

## Root cause
`allowed.Evaluate()` matched SANs against Go's parsed slices
(`DNSNames`/`IPAddresses`/`URIs`/`EmailAddresses`), which only cover GeneralName
tags `{1,2,6,7}` and silently drop `{0,3,4,5,8}` (otherName, x400Address,
directoryName, ediPartyName, registeredID). cert-manager copies the **raw** SAN
extension into the issued cert, so those dropped entries shipped unseen by policy.

## Fix
Parse the raw SAN extension (OID `2.5.29.17`) directly and evaluate every
GeneralName — **fail-closed**:
- tags `{1,2,6,7}` keep using the existing parsed-slice evaluators (unchanged);
- `otherName` (tag 0) is denied unless its OID is opted in via the new additive
  `allowed.otherNames` field (`oid` + `values`/`required`/`validations`);
- tags `{3,4,5,8}` are always denied (no opt-in).

DNS/IP/URI/email-only CSRs evaluate exactly as before.

The policy webhook also validates `allowed.otherNames` at **admission** (same as
every other `allowed.*` field): a `required` entry must declare `values`/
`validations`, every CEL rule must compile, and each `oid` must be a valid,
canonical dotted ASN.1 OID — so misconfigurations fail fast instead of producing
confusing runtime denials.

## Migration impact
| Scenario | After |
|---|---|
| DNS/IP/URI/email SANs only | works unchanged |
| SAN UPN `otherName` (AD / EAP-TLS / smart-card) | denied by default — add `allowed.otherNames: [{oid: "1.3.6.1.4.1.311.20.2.3", values: ["*@tenant.example.com"]}]` |
| SAN `directoryName`/`x400Address`/`ediPartyName`/`registeredID` | always denied, no opt-in (rare for cert-manager; Istio/SPIFFE/CSI use URI SANs, unaffected) |

Denied requests carry an `Approved=False` condition naming the offending SAN/OID
and the denying policy; the existing
`certmanager_approverpolicy_certificaterequests_approval{status="denied"}` metric
flags an upgrade-induced rise.


pkg/internal/approver/allowed/evaluator.go
```diff
@@ -104,12 +167,63 @@ type evaluator struct {
...
+// every encountered value decoded to a string — operators reading audit logs
+// should not see hex-encoded blobs and assume they were plain text.
+func (e evaluator) subjectValuesForOID(oid asn1.ObjectIdentifier) (values []string, allStrings bool) {
+	allStrings = true
+	for _, rdn := range e.subjectRDNs {
+		for _, atv := range rdn {
+			if !atv.Type.Equal(oid) {
+				continue
+			}
+			if s, ok := atv.Value.(string); ok {
```
inteon: Could you check in the cert-manager codebase, there should be similar code in there.
There might be some issue with the 'string' type vs other types.

pkg/internal/approver/allowed/evaluator_rawbypass_test.go
```diff
@@ -0,0 +1,313 @@
...
+				gen.CertificateRequest("", gen.SetCertificateRequestCSR(csr)))
+			assert.NoError(t, err)
+			assert.Equal(t, approver.ResultDenied, resp.Result,
+				"%s SAN entry must always be denied; got Result=%v Message=%q",
+				name, resp.Result, resp.Message)
+		})
+	}
+}
+
+// TestEvaluate_VC53795_SANUnaffectedByExistingFlow — a SAN containing only
```
SgtCoDFish: suggestion: Probably shouldn't refer to VC53795 in code here - it's not accessible to the vast majority of people and it's confusing.

mladen-rusev-cyberark: i will clean it up

pkg/internal/approver/allowed/evaluator.go
```diff
@@ -162,6 +201,81 @@ func (e evaluator) Usages() field.ErrorList {
 	return el
 }
 
+// OtherNames evaluates the SAN GeneralName entries that the parsed-slice
+// evaluators above (DNSNames/IPAddresses/URIs/EmailAddresses) do not cover.
+// otherName entries (tag 0) are permitted only when their OID is listed in
+// allowed.otherNames; directoryName, x400Address, ediPartyName and
+// registeredID entries are always denied — there is no opt-in for these.
+func (e evaluator) OtherNames() field.ErrorList {
+	fldPath := e.fldPath.Child("otherNames")
```
SgtCoDFish: suggestion(AI): I think the AI summary is good here - I've reviewed it.

I agree with the suggestion I think; I don't think otherNames should appear in the field path for say "X400Addresses" and similar.

---

Misleading field path for non-otherName SAN denials — evaluator.go:213-225

Denials for directoryName, x400Address, ediPartyName, and registeredID are all reported under spec.allowed.otherNames:

fldPath := e.fldPath.Child("otherNames")
...
el = append(el, field.Invalid(fldPath, "directoryName", "directoryName SAN entries are not permitted"))

But these types have nothing to do with the otherNames policy field — they're unconditionally denied. An operator reading "spec.allowed.otherNames: Invalid value: \"directoryName\": ..." will look at their otherNames: block and find nothing that could fix it. Suggestion:

sanPath := e.fldPath.Child("subjectAltName")   // or e.fldPath root
el = append(el, field.Invalid(sanPath.Child("directoryName"), ..., "..."))

This is the same concern Copilot flagged on the SAN-decode-failure message (already fixed) — worth applying consistently.

pkg/internal/approver/allowed/evaluator.go
```diff
@@ -59,10 +64,38 @@ func (a allowed) Evaluate(_ context.Context, policy *policyapi.CertificateReques
...
+			// Fail closed: a SAN extension we cannot fully decode must never
+			// be approved, since cert-manager would still sign its raw bytes.
+			return approver.EvaluationResponse{
+				Result:  approver.ResultDenied,
+				Message: fmt.Sprintf("subjectAltName (SAN) extension could not be decoded: %v", err),
+			}, nil
+		}
+		sans = parsed
+		// crypto/x509 rejects CSRs carrying duplicate extensions, so the
+		// first SAN extension is guaranteed to be the only one.
```
SgtCoDFish: suggestion(AI): Again I've reviewed this AI comment and I think I agree.

---

Duplicate-SAN-extension assumption is stdlib-dependent — evaluator.go:75-92

sans = parsed
// crypto/x509 rejects CSRs carrying duplicate extensions, so the
// first SAN extension is guaranteed to be the only one.
break

I confirmed this holds in the pinned Go version (crypto/x509/x509.go:2039 in Go 1.26.x, and go.mod requires go 1.26.0). But because this is security-critical and the failure mode is silent evaluator/signer divergence, it's worth belt-and-braces: count the SAN extensions first and deny if > 1. That way the code is safe even if Go's behavior changes, if the assumption breaks for a stub / mock CSR parser in tests, or if a downstream fork removes the check. Small cost, meaningful safety margin for a fix that is itself defending against divergence.

pkg/internal/approver/allowed/evaluator.go
```diff
@@ -59,10 +64,38 @@ func (a allowed) Evaluate(_ context.Context, policy *policyapi.CertificateReques
 		return approver.EvaluationResponse{}, err
 	}
 
+	// Decode the raw SAN extension with cert-manager's own SAN parser so the
+	// approver evaluates the exact GeneralName entries the signer will emit.
+	// crypto/x509 silently drops GeneralName tags outside {1,2,6,7}
```
SgtCoDFish, line 69: suggestion: I wouldn't mind seeing a unit test which asserts "crypto/x509 silently drops GeneralName tags outside {1,2,6,7} (otherName, x400Address, directoryName, ediPartyName, registeredID)"

That way, we'd see in a Go version upgrade if that behaviour changes because our tests would start failing.

Not a blocker - it's definitely gold-plating to do this. But I love unit testing assertions like that!

SgtCoDFish (changes requested):
Few nitty comments (2 AI, one human) which I think are more gold-plating than actual issues.

What do you think? Anything worth adding here?

SgtCoDFish (approved):
/lgtm
/approve

I can't see a reason not to merge. This is definitely an improvement!

PR #923, opened 2026-06-09
VC-53793: Add CEL cost limit

wallrj-cyberark:
## Summary

- Add `cel.CostLimit(1000000)` and `cel.InterruptCheckFrequency(100)` to CEL
  program options, preventing any single CEL expression from consuming more
  than ~0.1 seconds of CPU time. Expressions exceeding the limit are
  terminated with an "actual cost limit exceeded" error instead of running to
  completion.
- Add an addendum to the CEL design doc recording the change and rationale.

## Motivation

A deeply nested `.all()` comprehension (e.g. 8-deep `[0..9].all(...)` — 10^8
iterations) would hang the single reconcile worker
(`MaxConcurrentReconciles:1`) indefinitely, blocking all certificate approvals
cluster-wide.

`CertificateRequestPolicy` is cluster-scoped and should only be accessible to
admins, so a deliberate DoS is unlikely. However, a cost limit still provides
value as defense-in-depth against accidental complexity — a rule that looks
reasonable can have surprising evaluation cost.

The 1,000,000 cost unit budget is the de facto standard across the Kubernetes
ecosystem — not just the apiserver, but also Gatekeeper, Cluster API, Istio,
Cilium Tetragon, Flux, and Pinniped all use the same value. It corresponds to
roughly 0.1 seconds of evaluation time, which is generous for the simple
expressions typical of CertificateRequestPolicy rules. None of the existing
test expressions come close to hitting it.

## Upstream references

The `PerCallLimit` and `CheckFrequency` constants originate from the
Kubernetes apiextensions-apiserver CEL validation budget:

- kubernetes/kubernetes#108482 — introduced `cel.CostLimit` for CRD CEL validation
- kubernetes/kubernetes#109122 — halved the limits to current values (`PerCallLimit = 1000000`)
- kubernetes/kubernetes#115747 — extracted constants to [`k8s.io/apiserver/pkg/apis/cel/config.go`](https://github.com/kubernetes/kubernetes/blob/9e570c412469/staging/src/k8s.io/apiserver/pkg/apis/cel/config.go)

Other Kubernetes controllers using the same value:
- [OPA Gatekeeper](https://github.com/open-policy-agent/gatekeeper) — imports `celconfig.PerCallLimit`
- [Cluster API](https://github.com/kubernetes-sigs/cluster-api) — comments: "k/k CRD validation also uses celconfig.PerCallLimit"
- [Istio](https://github.com/istio/istio), [Cilium Tetragon](https://github.com/cilium/tetragon), [Flux](https://github.com/fluxcd/flux2), [Pinniped](https://github.com/vmware-tanzu/pinniped) — all use 1,000,000 / 100

Design context: [KEP for CRD Validation Expression Language](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/2876-crd-validation-expression-language)

## Test plan

- [x] `Test_Validator_Validate_CostLimitExceeded`: compiles the 8-deep nested
  comprehension PoC and verifies `Validate` returns a cost limit error in
  ~0.65s instead of hanging
- [x] All existing tests pass (`go test ./pkg/internal/approver/validation/`)

*Generated with Claude (Opus 4.6)*

### DIFF
--- a/pkg/internal/approver/validation/validator_test.go
+++ b/pkg/internal/approver/validation/validator_test.go
@@ -91,6 +91,32 @@ func Test_Validator_Validate(t *testing.T) {
 	}
 }
 
+// Test_Validator_Validate_CostLimitExceeded verifies that the CEL runtime
+// cost limit terminates expensive expressions before they hang the reconcile
+// worker. The expression uses 8-deep nested comprehensions (~10^8 iterations).
+//
+// If cel.CostLimit is accidentally removed, this test will hang rather than
+// fail cleanly — that is deliberate: the CI test timeout will catch it, and
+// the resulting stack trace pointing at program.Eval is more diagnostic than
+// a synthetic deadline error.
+func Test_Validator_Validate_CostLimitExceeded(t *testing.T) {
+	v := &validator{expression: `
+		[0,1,2,3,4,5,6,7,8,9].all(a,
+		[0,1,2,3,4,5,6,7,8,9].all(b,
+		[0,1,2,3,4,5,6,7,8,9].all(c,
+		[0,1,2,3,4,5,6,7,8,9].all(d,
+		[0,1,2,3,4,5,6,7,8,9].all(e,
+		[0,1,2,3,4,5,6,7,8,9].all(f,
+		[0,1,2,3,4,5,6,7,8,9].all(g,
+		[0,1,2,3,4,5,6,7,8,9].all(h, self != ""))))))))`}
+	err := v.compile()
+	assert.NoError(t, err, "expression must compile successfully")
+
+	_, err = v.Validate("anything", cmapi.CertificateRequest{})
+	assert.Error(t, err, "evaluation must be terminated by the cost limit")
+	assert.Contains(t, err.Error(), "cost limit exceeded")
+}
+
 func newCertificateRequest(namespace string) cmapi.CertificateRequest {
 	request := cmapi.CertificateRequest{}
 	request.SetNamespace(namespace)
--- a/design/20230726-cel-policy.md
+++ b/design/20230726-cel-policy.md
@@ -457,3 +457,29 @@ Let me start with a quote from the
 The last sentence describes the requirement that could need custom CEL function(s) to allow the user to specify allowed
 CSR attributes. We cannot expect the user to include validations for **all** CSR attributes in the expression(s). So I
 think we at least need a function to list allowed CSR attribute usage.
+
+## Addendum: CEL runtime cost limit (June 2025)
+
+A security review identified that a deeply nested comprehension (e.g. 8-deep
+`[0..9].all(...)` — 10^8 iterations) can hang the single reconcile worker
+indefinitely, blocking all certificate approvals cluster-wide. Even though
+`CertificateRequestPolicy` is cluster-scoped and restricted to admins,
+accidental complexity in a rule can trigger this.
+
+As a mitigation, **`cel.CostLimit(1000000)`** and
+**`cel.InterruptCheckFrequency(100)`** are now passed to `env.Program()`
+([#923](https://github.com/cert-manager/approver-policy/pull/923)). These
+match the Kubernetes apiextensions-apiserver `PerCallLimit` and
+`CheckFrequency` constants
+([`k8s.io/apiserver/pkg/apis/cel/config.go`](https://github.com/kubernetes/kubernetes/blob/9e570c412469/staging/src/k8s.io/apiserver/pkg/apis/cel/config.go)),
+which are the de facto standard across the ecosystem (also used by Gatekeeper,
+Cluster API, Istio, Cilium, Flux, and Pinniped). The budget of 1,000,000
+abstract cost units corresponds to roughly 0.1 seconds of evaluation time —
+generous for the simple expressions typical of `CertificateRequestPolicy`
+rules.
+
+The original design also noted that "we should probably consider adding some
+cache expiration logic to avoid appearing like a memory leak." The current
+implementation uses a global `sync.Map` with no eviction or expiry — every
+unique expression compiled during the lifetime of the controller remains in
+memory permanently. Cache lifecycle management is tracked separately.
--- a/pkg/internal/approver/validation/validator.go
+++ b/pkg/internal/approver/validation/validator.go
@@ -29,6 +29,24 @@ import (
 const (
 	varSelf    = "self"
 	varRequest = "cr"
+
+	// celCostLimit mirrors the Kubernetes apiextensions-apiserver PerCallLimit,
+	// measured in abstract cost units defined by the CEL cost model (e.g.,
+	// comparisons cost 1, list iteration costs 1 per element, string
+	// operations scale with length). Expressions exceeding this limit are
+	// terminated at runtime with an "actual cost limit exceeded" error
+	// instead of running to completion, preventing a malicious
+	// CertificateRequestPolicy from blocking the single reconcile worker
+	// indefinitely. See CWE-770.
+	// Upstream: https://github.com/kubernetes/kubernetes/blob/9e570c412469/staging/src/k8s.io/apiserver/pkg/apis/cel/config.go#L21
+	celCostLimit = 1000000
+
+	// celInterruptCheckFrequency controls how many comprehension iterations
+	// (.all(), .exists(), .map(), .filter()) elapse between cost-limit
+	// checks. Lower values catch breaches sooner but add overhead; higher
+	// values let the program overshoot the limit before being stopped.
+	// Upstream: https://github.com/kubernetes/kubernetes/blob/9e570c412469/staging/src/k8s.io/apiserver/pkg/apis/cel/config.go#L34
+	celInterruptCheckFrequency = 100
 )
 
 // Validator knows how to validate CSR attribute values in CertificateRequests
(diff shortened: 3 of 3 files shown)


wallrj-cyberark (commented):
Self-review of cert-manager/approver-policy#923.

*Generated with Claude (Opus 4.6)*

pkg/internal/approver/validation/cache.go
```diff
@@ -42,23 +57,26 @@ type cacheEntry struct {
 }
 
 func (c *cache) Get(expr string) (Validator, error) {
-	// First check if cache contains validator for expression
 	o, ok := c.m.Load(expr)
```
wallrj-cyberark, line 46: The two inline comments that were here previously have been removed:

1. `// First check if cache contains validator for expression` — restated what `c.m.Load(expr)` already says.
2. `// Expression did not exist in cache. Create a new validator, compile it and add the result to cache. // Theoretically this could lead to the same expression being compiled multiple times, but guarding against that would require locking and increase complexity.` — the first sentence restated the code. The "theoretically" observation is still true (and still acceptable) but the comment was entangled with the removed `LoadOrStore` line and would need rewriting to account for the new size-cap logic. Rather than rewrite it, I removed it: the behaviour is clear from the code and the `maxCacheEntries` doc comment on the `Cache` interface already explains the bounding semantics.

pkg/internal/approver/validation/validator_test.go
```diff
@@ -91,6 +91,26 @@ func Test_Validator_Validate(t *testing.T) {
 	}
 }
```
wallrj-cyberark, line 93: This test was run against the **unpatched `main` branch** (by copying it into the main worktree). It hangs for 30s then panics, confirming the vulnerability:

```
=== RUN   Test_Validator_Validate_CostLimitExceeded
panic: test timed out after 30s
    running tests:
        Test_Validator_Validate_CostLimitExceeded (30s)

goroutine 21 [runnable]:
github.com/google/cel-go/interpreter.(*evalAttr).Eval(...)
    ...cel-go@v0.27.0/interpreter/interpretable.go:1219
    [8 nested evalFold frames — one per comprehension depth]
github.com/cert-manager/approver-policy/...(*validator).Validate(...)
    .../validation/validator.go:99
FAIL  .../approver/validation  30.025s
```

On this branch, the same test passes in ~0.7s:
```
=== RUN   Test_Validator_Validate_CostLimitExceeded
--- PASS: Test_Validator_Validate_CostLimitExceeded (0.73s)
```

pkg/internal/approver/validation/cache.go
```diff
@@ -16,14 +16,28 @@ limitations under the License.
...
+	"sync/atomic"
+)
+
+const (
+	// maxCacheEntries bounds the number of compiled CEL programs held in
+	// memory. Without this cap, an attacker can inflate controller memory
+	// by submitting dry-run CREATEs with unique large rule strings, each
+	// of which is compiled and stored forever. See CWE-770.
+	maxCacheEntries = 1000
+)
```
wallrj-cyberark: Unlike the `celCostLimit` and `celInterruptCheckFrequency` constants in `validator.go`, this cache cap has no upstream Kubernetes equivalent.

### How Kubernetes handles compiled CEL programs

In Kubernetes, CEL validation expressions are defined in the CRD schema (`x-kubernetes-validations` rules), not in individual custom resources. When a CRD is created or updated, [`cel.NewValidator`](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apiserver/schema/cel/validation.go) eagerly compiles every expression into a tree of [`Validator` structs](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apiserver/schema/cel/validation.go) mirroring the schema structure. Each node holds its own `[]CompilationResult` (containing `cel.Program` objects) as a struct field. When a CR is validated, these pre-compiled programs are evaluated with the CR's data as input. When the CRD is deleted, the entire tree is garbage collected.

There is no separate cache or map — just struct fields on a per-CRD object. The [comment "CEL programs are compiled and cached here"](https://github.com/kubernetes/kubernetes/blob/9e570c412469/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/strategy.go#L78) in the Kubernetes source uses "cached" loosely to mean "compiled once and held in memory for the lifetime of the CRD's serving context." The compiled programs are stored on the [`customResourceStrategy`](https://github.com/kubernetes/kubernetes/blob/9e570c412469/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/strategy.go#L63) struct, which is [constructed per CRD version](https://github.com/kubernetes/kubernetes/blob/9e570c412469/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go#L835).

### Why approver-policy needs a cache cap

approver-policy uses a different architecture: instead of eagerly compiling per-CRP, a global `sync.Map` keyed by expression string lazily compiles on first encounter. This means the webhook populates the cache for every unique expression it sees, including `--dry-run=server` requests that never persist an object. Without a cap, an attacker can inflate controller memory indefinitely.

### Could approver-policy adopt the Kubernetes approach?

In principle, yes: compiled validators could be stored per-CertificateRequestPolicy (tied to the CRP lifecycle, like Kubernetes ties them to CRDs). This would eliminate the dry-run vector entirely. See #924 for a proof of concept demonstrating this approach. However, it is a significantly larger refactor:
- The current `validation.Cache` is a singleton on the [`allowed` approver](https://github.com/cert-manager/approver-policy/blob/087b527/pkg/internal/approver/allowed/allowed.go#L38-L41), shared by both the webhook and the reconciler. Splitting it into per-CRP storage would touch the approver interface boundary.
- Expression deduplication would be lost — if two CRPs share the same rule, it would be compiled twice.
- The webhook currently compiles lazily on first encounter. Moving to eager compilation at CRP create/update would change the error surface.

The fixed-cap approach in this PR is the minimal change that closes the vulnerability. A lifecycle-bound redesign could follow as a separate improvement.

*Generated with Claude (Opus 4.6)*

pkg/internal/approver/validation/cache.go
```diff
@@ -42,23 +57,26 @@ type cacheEntry struct {
...
-	o, _ = c.m.LoadOrStore(expr, &cacheEntry{validator: v, err: err})
+
+	if c.size.Load() >= maxCacheEntries {
+		return v, err
+	}
+
+	o, loaded := c.m.LoadOrStore(expr, &cacheEntry{validator: v, err: err})
+	if !loaded {
+		c.size.Add(1)
+	}
```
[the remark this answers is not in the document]
wallrj-cyberark: Added a code comment explaining this is a deliberate choice: the size check and store are intentionally non-atomic to avoid introducing a mutex on every cache access. The overshoot is bounded by the number of in-flight goroutines, which is acceptable for a safety-net cap.

*Generated with Claude (Opus 4.6)*

pkg/internal/approver/validation/validator_test.go
```diff
@@ -91,6 +91,26 @@ func Test_Validator_Validate(t *testing.T) {
...
+		[0,1,2,3,4,5,6,7,8,9].all(f,
+		[0,1,2,3,4,5,6,7,8,9].all(g,
+		[0,1,2,3,4,5,6,7,8,9].all(h, self != ""))))))))`}
+	err := v.compile()
+	assert.NoError(t, err, "expression must compile successfully")
+
+	_, err = v.Validate("anything", cmapi.CertificateRequest{})
+	assert.Error(t, err, "evaluation must be terminated by the cost limit")
+	assert.Contains(t, err.Error(), "cost limit exceeded")
+}
```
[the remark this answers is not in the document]
wallrj-cyberark: Added a doc comment to the test explaining the deliberate design: if `cel.CostLimit` is accidentally removed, this test hangs rather than failing cleanly. The CI test timeout catches it, and the resulting stack trace pointing at `program.Eval` is more diagnostic than a synthetic deadline error.

*Generated with Claude (Opus 4.6)*

pkg/internal/approver/validation/cache.go
```diff
@@ -42,23 +57,30 @@ type cacheEntry struct {
...
 	if err != nil {
 		v = nil
 	}
-	o, _ = c.m.LoadOrStore(expr, &cacheEntry{validator: v, err: err})
+
+	// The size check and store are deliberately non-atomic to avoid
+	// introducing a mutex on every cache access. Under concurrent load
+	// the map may briefly exceed maxCacheEntries by the number of
+	// in-flight goroutines, which is acceptable for a safety-net cap.
+	if c.size.Load() >= maxCacheEntries {
```
inteon: When the cache is full, this will just stop caching, I was expecting some kind of LRU mechanism.

wallrj-cyberark: Good point. In hindsight, combining the cache cap with the CEL cost limit in a single PR was a mistake — they are separate concerns and this made the PR harder to review.

The cache changes have been removed from this PR, which now contains only the CEL cost limit. Cache lifecycle management will be addressed separately: #924 has a PoC for per-CRP lifecycle-bound caching, and LRU is another option worth exploring (there is no LRU library in the current dependency tree, but it could be implemented with the standard library's `container/list` + `map`).

*Generated with Claude (Opus 4.6)*

wallrj-cyberark: See also #926 
 * #926

PR #924, opened 2026-06-09
VC-53793: Add per-CRP lifecycle-bound CEL validator cache

wallrj-cyberark:
## Summary

- Key compiled CEL validators by `(CertificateRequestPolicy name, expression)`,
  scoping cached programs to their owning policy.
- Populate the cache only when evaluating CertificateRequests against a
  persisted policy (`Cache.Get`). The admission webhook uses `Cache.Compile`,
  which checks that an expression compiles **without** caching it, so dry-run
  or subsequently-rejected requests — never persisted, so never cleaned up by a
  delete event — cannot leak entries.
- Discard a CRP's cached validators when it is deleted (via the
  `ValidatorCleaner` interface) and at the start of webhook `Validate()` on
  update.

## Motivation

The CEL validator cache uses a global `sync.Map` keyed by expression string
with no eviction — every unique expression compiled during the controller's
lifetime remains in memory permanently. The [CEL design doc](design/20230726-cel-policy.md)
flagged this: "we should probably consider adding some cache expiration logic
to avoid appearing like a memory leak."

This PR ties cache lifetime to the CRP lifecycle, analogous to how the
Kubernetes apiextensions-apiserver ties compiled CEL validators to the CRD
[`customResourceStrategy`](https://github.com/kubernetes/kubernetes/blob/9e570c412469/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/strategy.go#L63)
struct. See the `doc.go` file for detailed upstream links.

## Design

| Aspect | Before (global cache) | After (per-CRP lifecycle) |
|---|---|---|
| Key | Expression string | (CRP name, expression) |
| Eviction | None | On CRP delete or update (best-effort) |
| Admission checks | Cached permanently | Compiled, not stored (`Cache.Compile`) |
| Memory bound | Unbounded | Live policies + bounded eviction tail |
| Deduplication | Shared across CRPs | Compiled per CRP |

Losing expression deduplication across CRPs is acceptable — CRPs are
admin-managed, low-churn resources, and the number of unique expressions per
cluster is small.

Eviction is **best-effort and correctness-independent**: because the expression
string is part of the cache key, a lookup only ever returns the validator for
that exact expression, so a stale entry is never returned for the wrong input.
In a multi-replica deployment the webhook `Remove` runs only on the replica that
served the admission and delete cleanup runs only on the leader, so other
replicas may retain old entries until restart — a bounded memory cost, never a
wrong-answer risk.

## Test plan

- [x] `Test_Cache_Get`: basic caching and error caching, scoped by policy name
- [x] `Test_Cache_Compile`: `Compile` returns a validator (or error) without
  caching — repeated calls yield distinct validators
- [x] `Test_Cache_Compile_DoesNotPopulate`: a later `Get` does not see anything
  left by `Compile`, so admission never populates the lifecycle cache
- [x] `Test_Cache_Get_IsolatedByOwner`: same expression under different owners
  produces separate entries
- [x] `Test_Cache_Remove`: after Remove, Get compiles a fresh validator
- [x] `Test_Cache_Remove_OnlyAffectsTarget`: Remove does not affect other
  owners
- [x] `Test_Cache_Remove_ThenRepopulate`: simulates CRP update — Remove clears
  old entries, subsequent Get compiles fresh
- [x] `make test` and `verify-golangci-lint` pass locally; all existing tests
  pass

### DIFF
--- a/design/20230726-cel-policy.md
+++ b/design/20230726-cel-policy.md
@@ -479,7 +479,46 @@ generous for the simple expressions typical of `CertificateRequestPolicy`
 rules.
 
 The original design also noted that "we should probably consider adding some
-cache expiration logic to avoid appearing like a memory leak." The current
-implementation uses a global `sync.Map` with no eviction or expiry — every
-unique expression compiled during the lifetime of the controller remains in
-memory permanently. Cache lifecycle management is tracked separately.
+cache expiration logic to avoid appearing like a memory leak." The initial
+implementation used a global `sync.Map` keyed by expression string with no
+eviction or expiry — every unique expression compiled during the lifetime of
+the controller remained in memory permanently.
+
+## Addendum: per-CRP lifecycle-bound validator cache (June 2026)
+
+[#924](https://github.com/cert-manager/approver-policy/pull/924) replaced the
+global expression-keyed cache with a two-level cache keyed by
+`(CertificateRequestPolicy name, expression)`. This ties compiled CEL
+programs to the lifecycle of their owning CRP, mirroring how the Kubernetes
+apiextensions-apiserver stores compiled validators on the
+[`customResourceStrategy`](https://github.com/kubernetes/kubernetes/blob/9e570c412469/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/strategy.go#L63)
+struct per CRD version.
+
+Cache entries are managed as follows:
+
+- **During evaluation**: `Get(policyName, expression)` compiles on first use and
+  caches the program under the persisted policy's name. This is the only path
+  that populates the cache.
+- **At admission (create/update)**: the webhook `Validate()` calls `Compile`,
+  which checks that each expression compiles **without** caching it, then calls
+  `Remove(policyName)` to drop entries left by the previous generation. Because
+  admission never populates the cache, dry-run and subsequently-rejected
+  requests — which are never persisted and so never receive a delete event —
+  cannot leak entries.
+- **On CRP deletion**: the controller calls `ValidatorCleaner.CleanupValidators`,
+  which removes all cached entries for that policy.
+
+Eviction is **best-effort**. The expression string is part of the cache key, so
+a stale entry can never be returned for the wrong input — correctness never
+depends on eviction; it only bounds memory. In a multi-replica deployment each
+admission request is served by a single replica and the delete reconciler runs
+only on the elected leader, so old entries for an updated or deleted policy may
+linger on other replicas until the process restarts. Steady-state memory is
+therefore bounded by the live policies plus a bounded tail of not-yet-evicted
+entries, rather than growing with every unique expression ever seen as the
+previous global cache did.
+
+A fully replica-complete eviction would require evicting on the delete webhook
+path (which runs on every replica) rather than in the leader-only controller;
+this was judged unnecessary given eviction is correctness-irrelevant and the
+remaining growth is bounded.
--- a/pkg/internal/approver/allowed/validation.go
+++ b/pkg/internal/approver/allowed/validation.go
@@ -36,6 +36,17 @@ func (a allowed) Validate(_ context.Context, policy *policyapi.CertificateReques
 		}, nil
 	}
 
+	// On CREATE/UPDATE, discard any validators cached under this policy name by
+	// a previous generation so an updated CRP does not retain compiled programs
+	// for expressions it no longer declares. The expressions below are checked
+	// with Compile (not Get) so this admission path never repopulates the
+	// lifecycle cache: a dry-run or subsequently-rejected request must not leave
+	// entries behind. The cache is repopulated lazily during evaluation of the
+	// persisted policy. Remove is best-effort per replica (see the cache docs).
+	if policy.Name != "" {
+		a.validators.Remove(policy.Name)
+	}
+
 	var (
 		el      field.ErrorList
 		allowed = policy.Spec.Allowed
@@ -83,7 +94,7 @@ func (a allowed) Validate(_ context.Context, policy *policyapi.CertificateReques
 				}
 			}
 			for i, validation := range stringSlice.slice.Validations {
-				if _, err := a.validators.Get(validation.Rule); err != nil {
+				if _, err := a.validators.Compile(validation.Rule); err != nil {
 					el = append(el, field.Invalid(stringSlice.path.Child("validations").Index(i), validation.Rule, err.Error()))
 				}
 			}
@@ -98,7 +109,7 @@ func (a allowed) Validate(_ context.Context, policy *policyapi.CertificateReques
 				}
 			}
 			for i, validation := range stringI.string.Validations {
-				if _, err := a.validators.Get(validation.Rule); err != nil {
+				if _, err := a.validators.Compile(validation.Rule); err != nil {
 					el = append(el, field.Invalid(stringI.path.Child("validations").Index(i), validation.Rule, err.Error()))
 				}
 			}
--- a/pkg/internal/approver/validation/cache.go
+++ b/pkg/internal/approver/validation/cache.go
@@ -16,24 +16,49 @@ limitations under the License.
 
 package validation
 
-import "sync"
+import (
+	"sync"
+)
 
-// Cache maintains a cache of compiled validators.
-// The current implementation is a simple lazy cache meaning:
+// Cache maintains compiled CEL validators keyed by CertificateRequestPolicy
+// name and expression string, analogous to how the Kubernetes
+// apiextensions-apiserver ties compiled validators to the CRD
+// customResourceStrategy (see package doc for details and upstream links).
 //
-// 1. Whenever a validator is requested, it first checks the cache.
-// 2. If a compiled validator exists for the supplied CEL expression, it is returned.
-// 3. If the validator doesn't exist in the cache, a new validator is created, compiled, added to the cache, and returned.
+// Compiled programs are tied to the CRP lifecycle: populated on demand during
+// evaluation via [Cache.Get] and discarded via [Cache.Remove] when the CRP is
+// updated or deleted. Eviction is best-effort: because the expression string
+// is part of the cache key, a stale entry can never be returned for the wrong
+// input, so correctness never depends on it — Remove only bounds memory.
+//
+// CertificateRequestPolicy is cluster-scoped so the name is globally unique.
 type Cache interface {
-	// Get returns a compiled validator for the supplied CEL expression.
-	// Any compilation errors will be returned to the caller.
-	//
-	// The supplied CEL expression must output a bool.
-	Get(expr string) (Validator, error)
+	// Get returns a compiled validator for the given CEL expression, cached
+	// for the lifetime of the named CertificateRequestPolicy. policyName must
+	// be non-empty; callers that only need to check whether an expression
+	// compiles, without retaining it, should use [Cache.Compile] instead.
+	Get(policyName string, expr string) (Validator, error)
+
+	// Compile compiles the CEL expression and returns the validator without
+	// caching it. It is used for admission-time validity checks, where the
+	// result is transient and must not grow the lifecycle cache — for example
+	// dry-run or rejected CREATE requests, which are never persisted and so
+	// would otherwise leak entries that no delete event ever cleans up.
+	Compile(expr string) (Validator, error)
+
+	// Remove discards all compiled validators for the named
+	// CertificateRequestPolicy. It is best-effort memory cleanup (see the type
+	// documentation); correctness does not depend on it.
+	Remove(policyName string)
 }
 
 type cache struct {
-	m sync.Map
+	m sync.Map // string (CRP name) -> *crpEntry
+}
+
+type crpEntry struct {
+	mu         sync.Mutex
+	validators map[string]*cacheEntry
 }
 
 type cacheEntry struct {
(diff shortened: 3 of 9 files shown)


pkg/internal/approver/validation/cache_test.go
```diff
@@ -22,35 +22,131 @@ import (
...
+	c.Remove("policy-1")
+
+	// Re-populate with only expression 'a', as if the updated CRP dropped 'b'.
+	v1aNew, err := c.Get("policy-1", "self == 'a'")
+	assert.NoError(t, err)
+
+	// NotSame asserts the two values have different pointer addresses,
+	// proving a fresh validator was compiled rather than returning the old one.
+	assert.NotSame(t, v1a, v1aNew, "after Remove+re-Get, validator must be freshly compiled")
+}
```
wallrj-cyberark, line 172: **CI blocker.** `pull-cert-manager-approver-policy-verify` fails here:

```
pkg/internal/approver/validation/cache_test.go:153:1: File is not properly formatted (gci)
```

Run the repo's lint autofix before pushing, e.g. `_bin/tools/golangci-lint run --fix ./pkg/internal/approver/validation/...` (or `make verify-golangci-lint` to reproduce). The build will stay red until this is resolved.

pkg/internal/approver/allowed/validation.go
```diff
@@ -83,7 +87,7 @@ func (a allowed) Validate(_ context.Context, policy *policyapi.CertificateReques
 				}
 			}
 			for i, validation := range stringSlice.slice.Validations {
-				if _, err := a.validators.Get(validation.Rule); err != nil {
+				if _, err := a.validators.Get(policy.Name, validation.Rule); err != nil {
```
wallrj-cyberark: This (and the matching call at line 105) populates the per-CRP cache under `policy.Name` at **admission time**, before the object is persisted. Two leak paths follow:

- An explicitly-named **dry-run** CREATE has a non-empty `policy.Name`, so its compiled programs are cached but the object is never persisted → no delete event → the entry leaks.
- A CREATE whose rules compile but is then **rejected** (here or by a later webhook) leaves the successfully-compiled rules cached under a name that never gets cleaned up.

Since `Validate()` only needs to know whether each expression *compiles*, consider making these checks compile-only:

```go
if _, err := a.validators.Get("", validation.Rule); err != nil {
```

Keep `Remove(policy.Name)` at line 39 for update invalidation (it must run on the webhook path so every replica invalidates — see the controller comment). With this change, admission never populates the lifecycle cache; only real evaluations of persisted, Ready CRPs do, and those names always get delete cleanup.

pkg/internal/approver/validation/cache.go
```diff
@@ -16,51 +16,79 @@ limitations under the License.
...
 }
 
-func (c *cache) Get(expr string) (Validator, error) {
-	// First check if cache contains validator for expression
-	o, ok := c.m.Load(expr)
-	if ok {
-		ce := o.(*cacheEntry)
-		return ce.validator, ce.err
+func (c *cache) Get(owner string, expr string) (Validator, error) {
+	if owner == "" {
```
wallrj-cyberark: The empty-`owner` branch is documented as the dry-run guard, but empty owner is not equivalent to dry-run:

- A normal CREATE using `generateName` has an empty name at admission yet **is** persisted — it would be safe to cache, and skipping it here just defers the first compile to evaluation time (harmless).
- An explicitly-named **dry-run** CREATE has a non-empty name and so falls through to the caching path (see `validation.go:90`) — exactly the case this is meant to protect against.

The robust signal is the admission `dryRun` flag, which isn't plumbed down to this layer today. Either thread it through, or lean on making `Validate()` compile-only (see the `validation.go` comment) so this branch is only ever exercised by the validity check rather than relied upon as the dry-run defence.

wallrj-cyberark (commented):
## Review Summary

The lifecycle-bound design is a clear improvement on the unbounded global cache, the parallel with `customResourceStrategy` is well drawn, and the test coverage for the cache primitives is thorough. Three things are worth addressing before this leaves Draft:

1. **CI is red.** `pull-cert-manager-approver-policy-verify` fails on a `gci` formatting issue in `cache_test.go`. Run the repo formatter and force-push.

2. **The dry-run protection does not close the vector it claims to.** `Validate()` populates the per-CRP cache at admission time keyed by `policy.Name`. An explicitly-named dry-run CREATE — or any CREATE that compiles valid rules but is then rejected — caches compiled programs under a name that will never receive a delete event, so those entries leak. The empty-`owner` heuristic only skips caching for `generateName` creates (which *are* persisted and so are safe), and misses the named-dry-run case (which is *not* persisted). That is close to the opposite of the stated intent. Mitigation: CRPs are cluster-scoped and admin-RBAC-gated, so the blast radius is limited, but for a CWE-770 claim this should be tight. Simplest fix: make the validity check in `Validate()` compile-only (`Get("", …)`) and keep `Remove(policy.Name)` purely for update invalidation — admission then never populates the lifecycle cache; only real evaluations of persisted, Ready CRPs do.

3. **Deletion cleanup is leader-only.** `CleanupValidators` is driven from the controller, which in a multi-replica deployment runs only on the elected leader. Update invalidation goes through the webhook `Validate()` (all replicas), but deletion does not, so non-leader replicas retain entries for deleted CRPs indefinitely. Still better than before, but the doc's "bounded by the number of persisted CertificateRequestPolicies" is only true on the leader.

Remaining notes are minor (allocation on the hot path, `owner`/`policyName` naming, an addendum date). Detail inline.

*Review conducted with assistance from Claude (Opus 4.8)*

pkg/internal/approver/validation/doc.go
```diff
@@ -0,0 +1,54 @@
...
+//     number of CRDs in the cluster.
+//
+// approver-policy differs in that CEL expressions are opaque string values
+// inside CertificateRequestPolicy custom resources, not part of a CRD schema.
+// The controller compiles them lazily on first encounter rather than eagerly
+// at CRP create/update time. However, the lifecycle-bound storage principle
+// is the same: compiled programs are owned by their CRP and cleaned up on
+// deletion.
+//
+// Dry-run CREATE requests (where the policy has not been persisted and
```
wallrj-cyberark: "where the policy has not been persisted and therefore has no name" conflates two distinct things. A `generateName` create has no name yet **is** persisted; an explicitly-named dry-run create has a name and is **not** persisted. As written the empty-name guard misses the named dry-run case, so the CWE-770 claim in the next sentence is stronger than the implementation currently delivers. Please reword once the approach in `validation.go`/`cache.go` is settled.

pkg/internal/controllers/certificaterequestpolicies.go
```diff
@@ -167,7 +168,15 @@ func (c *certificaterequestpolicies) reconcileStatusPatch(ctx context.Context, r
 
 	policy := new(policyapi.CertificateRequestPolicy)
 	if err := c.client.Get(ctx, req.NamespacedName, policy); err != nil {
-		return reconcile.Result{}, nil, client.IgnoreNotFound(err)
+		if apierrors.IsNotFound(err) {
```
wallrj-cyberark, line 171: Deletion cleanup runs only from the controller, which under leader election executes on a single replica. The webhook serves admission on **all** replicas, so update invalidation (`Validate` → `Remove`) is replica-complete, but deletion cleanup is not — non-leader replicas keep entries for deleted CRPs until the process restarts.

To make deletion cleanup replica-complete, the natural mirror is to also evict on the delete webhook path (`ValidateDelete` currently no-ops without consulting the approvers). If that's out of scope, please soften the doc's "bounded by the number of persisted CertificateRequestPolicies" to reflect that deleted-CRP eviction is best-effort per replica.

Aside: this now uses `c.client.Get` (direct read) rather than the informer cache, which makes the `IsNotFound` → cleanup decision authoritative and avoids spurious eviction from a lagging cache — good.

pkg/internal/approver/validation/cache.go
```diff
@@ -16,51 +16,79 @@ limitations under the License.
...
+		v := &validator{expression: expr}
+		err := v.compile()
+		if err != nil {
+			return nil, err
+		}
+		return v, nil
+	}
+
+	entry, _ := c.m.LoadOrStore(owner, &crpEntry{
```
wallrj-cyberark: Minor: `LoadOrStore` evaluates its second argument on every call, so a fresh `crpEntry` and `map` are allocated and discarded on every cache **hit** too. On the evaluation hot path consider `Load` first, falling back to `LoadOrStore` only on miss:

```go
v, ok := c.m.Load(owner)
if !ok {
    v, _ = c.m.LoadOrStore(owner, &crpEntry{validators: make(map[string]*cacheEntry)})
}
ce := v.(*crpEntry)
```

pkg/internal/approver/validation/cache.go
```diff
@@ -16,51 +16,79 @@ limitations under the License.
...
 type Cache interface {
-	// Get returns a compiled validator for the supplied CEL expression.
-	// Any compilation errors will be returned to the caller.
+	// Get returns a compiled validator for the given CEL expression, scoped
+	// to the named CertificateRequestPolicy.
 	//
-	// The supplied CEL expression must output a bool.
-	Get(expr string) (Validator, error)
+	// If owner is empty, the expression is compiled but not cached.
+	Get(owner string, expr string) (Validator, error)
```
wallrj-cyberark: Nit: the parameter is `owner` here but every caller passes `policy.Name` and the cleanup method calls it `policyName`. Picking one name (the doc says owner *is* the CRP name) would make the data flow easier to grep.

design/20230726-cel-policy.md
```diff
@@ -479,7 +479,30 @@ generous for the simple expressions typical of `CertificateRequestPolicy`
...
-implementation uses a global `sync.Map` with no eviction or expiry — every
-unique expression compiled during the lifetime of the controller remains in
-memory permanently. Cache lifecycle management is tracked separately.
\ No newline at end of file
+cache expiration logic to avoid appearing like a memory leak." The initial
+implementation used a global `sync.Map` keyed by expression string with no
+eviction or expiry — every unique expression compiled during the lifetime of
+the controller remained in memory permanently.
+
+## Addendum: per-CRP lifecycle-bound validator cache (June 2025)
```
wallrj-cyberark: Nit: this addendum is dated "(June 2025)" but the PR is from June 2026. The earlier addendum at line 461 also says June 2025, so please double-check which is intended — if that one is genuinely from last year, this new one should read June 2026.

pkg/internal/approver/validation/cache_test.go
wallrj-cyberark: Fixed in a separate session — the `gci` formatting was corrected and the `verify` job is now green.

pkg/internal/approver/allowed/validation.go
wallrj-cyberark: Addressed in 2865d9a. `Validate()` now calls the new `Cache.Compile()` (compile-only) for both validity checks, so the admission path never populates the lifecycle cache — named dry-run and subsequently-rejected creates can no longer leak entries. `Remove(policy.Name)` is retained for update invalidation.

pkg/internal/approver/validation/cache.go
wallrj-cyberark: Addressed in 2865d9a. The empty-owner sentinel is gone. Admission validity checks use `Cache.Compile(expr)` (never cached); `Get(policyName, expr)` always caches and is used only by evaluation of persisted policies. The real `dryRun` flag is not relied upon — it isn't plumbed to this layer, and the compile-only path makes it unnecessary.

pkg/internal/approver/validation/doc.go
wallrj-cyberark: Reworded in 2865d9a. `doc.go` now describes the admission path as compile-only and the lifecycle cache as populated only during evaluation, keyed by `(name, expr)`, with the memory bound scoped to live policies plus a bounded eviction tail rather than claiming the empty-name guard closes CWE-770.

pkg/internal/controllers/certificaterequestpolicies.go
wallrj-cyberark: Acknowledged and documented rather than re-architected. Eviction is best-effort and correctness-independent: because the expression is part of the cache key, a stale entry is never returned for the wrong input. This is now stated in `doc.go`, this controller block, the `ValidatorCleaner` interface, and the design-doc addendum — which also records the all-replica delete-webhook option and why it was judged unnecessary.

pkg/internal/approver/validation/cache.go
wallrj-cyberark: Addressed in 2865d9a. `Get` now does a `Load` first and only `LoadOrStore`s on a miss, so cache hits no longer allocate a throwaway `crpEntry`/map.

pkg/internal/approver/validation/cache.go
wallrj-cyberark: Addressed in 2865d9a. `owner` is renamed to `policyName` across the cache API to match the callers and `CleanupValidators`.

design/20230726-cel-policy.md
wallrj-cyberark: Fixed in 2865d9a — the addendum is re-dated to June 2026.

wallrj-cyberark:
For reviewers: #926 is an alternative implementation that bounds the same cache with a fixed-size LRU (`k8s.io/utils/lru`) instead of tying entries to the CertificateRequestPolicy lifecycle. It is a smaller, self-contained change (one production file, `cache.go`) and carries no controller/webhook plumbing or multi-replica eviction caveats, at the cost of a fixed capacity rather than working-set-sized memory. Both close the memory-exhaustion variant of VC-53793; only one should be merged.

wallrj-cyberark:
Closing in favour of a simpler LRU approach. 
 * #926

PR #925, opened 2026-06-09
VC-53793: Replace recursive wildcard matching with iterative algorithm

wallrj-cyberark:
## Summary

- Replace recursive `matchRunes` with an iterative backtrack-checkpoint algorithm that runs in O(n*m), fixing exponential backtracking with multiple wildcards
- Add `Test_WildcardMatches_BacktrackingDoesNotHang` regression test

## Motivation

This hardens approver-policy against [CWE-770](https://cwe.mitre.org/data/definitions/770.html) (allocation of resources without limits or throttling). A security assessment identified two DoS variants in approver-policy:

1. **CEL comprehension DoS** — fixed in cert-manager/approver-policy#923
2. **Wildcard backtracking DoS** — fixed in this PR

The `matchRunes` function (originally adapted from [minio](https://github.com/minio/minio/blob/RELEASE.2020-06-22T03-12-50Z/pkg/wildcard/match.go)) used recursive backtracking to match `*` wildcards. With multiple wildcards in a pattern (e.g. `*a*a*a*a*a*a*a*a*!`), the worst case is O(2^n) — a pattern with 8 wildcards against a 50-character string explores ~2^50 branches, hanging the reconcile worker indefinitely.

The exponential cost is driven by the number of `*` in the **pattern**, which comes from a `CertificateRequestPolicy` (`allowed.*.values` and `selector.issuerRef` fields) — a privileged, cluster-scoped resource. The realistic risk is therefore an administrator **inadvertently** configuring a pathological pattern, rather than an attacker authoring a malicious CRP. Note the matched string comes from `CertificateRequest`s (a far less privileged, namespaced operation), so once such a pattern exists, a low-privileged requester can supply crafted values that hit the worst case — escalating a misconfiguration into a cluster-wide approval DoS.

The fix replaces the recursive algorithm with an iterative "backtrack checkpoint" technique: when a `*` is encountered, the pattern and string positions are recorded as a checkpoint. On mismatch, execution backtracks to the checkpoint (advancing the string position by one) instead of forking into two recursive branches.

## Correctness / prior art

This is not a bespoke algorithm — it is the standard **greedy wildcard matcher with a single backtracking checkpoint** (the "two-pointer" linear glob matcher). On each `*` we record only the most recent checkpoint (`starPx`/`starSx`) and, on a later mismatch, rewind the text pointer one past that checkpoint rather than forking. It runs in O(n*m); only the most recent `*` ever needs remembering.

References that establish correctness and the security motivation:

- **Russ Cox, ["Glob Matching Can Be Simple And Fast Too"](https://research.swtch.com/glob)** — presents the same checkpoint algorithm, explains why recursive matching is exponential (`n^e` for `e` stars), and frames it as a DoS class with real CVEs (CVE-2001-1501, CVE-2010-2632). Notes the same technique backs Go's `path/filepath.Match`, glibc's `glob(3)` and vsftpd.
- **Wikipedia, ["Matching wildcards"](https://en.wikipedia.org/wiki/Matching_wildcards)** (non-recursive section) — "only one such saved set [of pointers] is needed since just one successful match is required"; attributes iterative implementations to Dogan Kurt and Kirk J. Krauss (Dr. Dobb's, 2008).
- **[LeetCode #44 "Wildcard Matching"](https://leetcode.com/problems/wildcard-matching/)** — the canonical accepted O(m·n) greedy solution is structurally identical to this implementation.

**Note on upstream minio:** minio never fixed this. Both the originally-copied version and the current [`minio/pkg`](https://github.com/minio/pkg/blob/master/wildcard/match.go) still use recursive backtracking and remain vulnerable. We deliberately diverge: unlike minio we support only `*` (not `?`) and match on runes rather than bytes, so this matcher should **not** be re-synced from upstream. This is documented in the file header.

## Test plan

- [x] All existing `Test_WildcardSubset`, `Test_WildcardContains`, `Test_WildcardMatches` tests pass — the fix is internal to `matchRunes` only
- [x] `Test_WildcardMatches_BacktrackingDoesNotHang` completes instantly and fails fast (bounded 5s `select`) if the exponential algorithm is ever reinstated, rather than hanging until the global CI timeout
- [x] `make verify` passes (incl. `golangci-lint`)

### DIFF
--- a/pkg/internal/util/wildcard.go
+++ b/pkg/internal/util/wildcard.go
@@ -15,7 +15,14 @@ limitations under the License.
 */
 
 // !!
-// Modified from https://github.com/minio/minio/blob/RELEASE.2020-06-22T03-12-50Z.hotfix/pkg/wildcard/match.go
+// Originally adapted from minio's wildcard matcher (minio/minio, now archived):
+// https://github.com/minio/minio/blob/RELEASE.2020-06-22T03-12-50Z/pkg/wildcard/match.go
+//
+// matchRunes (below) has since been rewritten to an iterative algorithm and is
+// intentionally not a drop-in copy of the version now maintained at
+// https://github.com/minio/pkg/blob/main/wildcard/match.go: that one is still
+// recursive, and unlike minio we support only '*' (not '?') and match on runes
+// rather than bytes. Prefer not to re-sync from upstream.
 
 package util
 
@@ -65,27 +72,47 @@ func WildcardMatches(pattern, str string) bool {
 	return matchRunes([]rune(pattern), []rune(str))
 }
 
-// matchRunes will return whether the given rune slice matches the given
-// pattern using wildcards ('*').
+// matchRunes returns whether str matches pattern using '*' wildcards.
+// It uses an iterative backtrack-checkpoint algorithm that runs in O(n*m)
+// time, replacing the original recursive implementation which was O(2^n)
+// with multiple wildcards. See CWE-770.
+//
+// This is the standard greedy wildcard matcher: on each '*' we record a single
+// checkpoint (starPx/starSx) and, on a later mismatch, rewind str one past that
+// checkpoint rather than forking into two recursive calls. Only the most recent
+// '*' needs remembering. The same technique backs Go's path/filepath.Match,
+// glibc's glob(3) and many other production matchers. References:
+//   - Russ Cox, "Glob Matching Can Be Simple And Fast Too" https://research.swtch.com/glob
+//   - https://en.wikipedia.org/wiki/Matching_wildcards (non-recursive section)
 func matchRunes(pattern, str []rune) bool {
-	for len(pattern) > 0 {
-		switch pattern[0] {
+	px, sx := 0, 0
+	starPx, starSx := -1, -1
 
-		// If '*' then branch with recursive check for before and after '*'
-		case '*':
-			return matchRunes(pattern[1:], str) || (len(str) > 0 && matchRunes(pattern, str[1:]))
-
-		// If still strings to match or patter and string don't match at index, no match.
+	for sx < len(str) {
+		switch {
+		case px < len(pattern) && pattern[px] == '*':
+			// Record a checkpoint at the '*' and try to match it against the
+			// empty string first.
+			starPx, starSx = px, sx
+			px++
+		case px < len(pattern) && pattern[px] == str[sx]:
+			// Literal match; advance both positions.
+			px++
+			sx++
+		case starPx >= 0:
+			// Mismatch, but a '*' is still open: backtrack to it and let it
+			// consume one more character of str.
+			starSx++
+			sx = starSx
+			px = starPx + 1
 		default:
-			if len(str) == 0 || str[0] != pattern[0] {
-				return false
-			}
+			// Mismatch with no '*' to fall back on.
+			return false
 		}
-
-		str = str[1:]
-		pattern = pattern[1:]
 	}
 
-	// If both empty, then match
-	return len(str) == 0 && len(pattern) == 0
+	for px < len(pattern) && pattern[px] == '*' {
+		px++
+	}
+	return px == len(pattern)
 }
--- a/pkg/internal/util/wildcard_test.go
+++ b/pkg/internal/util/wildcard_test.go
@@ -18,7 +18,9 @@ package util
 
 import (
 	"fmt"
+	"strings"
 	"testing"
+	"time"
 )
 
 func Test_WildcardSubset(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/internal/util/wildcard.go
```diff
@@ -15,7 +15,15 @@ limitations under the License.
...
 
 // !!
-// Modified from https://github.com/minio/minio/blob/RELEASE.2020-06-22T03-12-50Z.hotfix/pkg/wildcard/match.go
+// Originally adapted from https://github.com/minio/minio/blob/RELEASE.2020-06-22T03-12-50Z.hotfix/pkg/wildcard/match.go
+//
+// The matcher has since been rewritten (see matchRunes below). The upstream
+// minio implementation — including the current version in
+// https://github.com/minio/pkg/blob/master/wildcard/match.go — still uses
+// recursive backtracking and remains vulnerable to the exponential
+// worst case fixed here (CWE-770 / VC-53793). We deliberately do not track
```
erikgb: I don't like references to things that are not publicly available. Not in the PR description, and at least not in the code. So I suggest removing VC-53793, which is an internal/closed Palo Alto reference.

wallrj-cyberark: Good point, thanks — agreed that closed/internal references don't belong in a public repository. I've removed the `VC-53793` reference from the code, the PR description, and the commit message, keeping the public [CWE-770](https://cwe.mitre.org/data/definitions/770.html) reference instead.

pkg/internal/util/wildcard_test.go
```diff
@@ -304,3 +306,29 @@ func Test_WildcardMatches(t *testing.T) {
...
 }
+
+// Test_WildcardMatches_BacktrackingDoesNotHang verifies that patterns with
+// multiple wildcards do not cause exponential backtracking. The pattern
+// "*a*a*a*a*a*a*a*a*!" against a string of 50 'a' characters would explore
+// ~2^50 branches with the old recursive algorithm; the iterative replacement
+// handles it in O(n*m). If the iterative algorithm is accidentally reverted,
+// the call below will not return and the bounded select fails the test with an
+// actionable message, rather than hanging until the global test timeout fires.
+func Test_WildcardMatches_BacktrackingDoesNotHang(t *testing.T) {
```
wallrj-cyberark, line 317: **Test evidence — this guard was verified against the previous implementation.**

I ran this exact test against the old recursive `matchRunes` from `main` (keeping the new test file, swapping only the source), then against the new iterative implementation:

```console
# Old recursive implementation (git restore --source=main -- pkg/internal/util/wildcard.go)
$ go test ./pkg/internal/util/ -run Test_WildcardMatches_BacktrackingDoesNotHang -v
=== RUN   Test_WildcardMatches_BacktrackingDoesNotHang
    wildcard_test.go:332: WildcardMatches("*a*a*a*a*a*a*a*a*!", ...) did not return within 5s; exponential backtracking has regressed
--- FAIL: Test_WildcardMatches_BacktrackingDoesNotHang (5.00s)
FAIL

# New iterative implementation
$ go test ./pkg/internal/util/ -run Test_WildcardMatches_BacktrackingDoesNotHang -v
=== RUN   Test_WildcardMatches_BacktrackingDoesNotHang
--- PASS: Test_WildcardMatches_BacktrackingDoesNotHang (0.00s)
PASS
```

So the guard fails fast (5.00s, via the bounded `select`) on the vulnerable code and passes instantly on the fix — confirming it would catch a regression rather than silently hanging CI.

pkg/internal/util/wildcard.go
```diff
@@ -65,27 +73,47 @@ func WildcardMatches(pattern, str string) bool {
...
+// with multiple wildcards. See CWE-770.
+//
+// This is the standard greedy wildcard matcher: on each '*' we record a single
+// checkpoint (starPx/starSx) and, on a later mismatch, rewind str one past that
+// checkpoint rather than forking into two recursive calls. Only the most recent
+// '*' needs remembering. The same technique backs Go's path/filepath.Match,
+// glibc's glob(3) and many other production matchers. References:
+//   - Russ Cox, "Glob Matching Can Be Simple And Fast Too" https://research.swtch.com/glob
+//   - https://en.wikipedia.org/wiki/Matching_wildcards (non-recursive section)
 func matchRunes(pattern, str []rune) bool {
```
wallrj-cyberark, line 87: For reviewers: here is an interactive, step-through animation of this `matchRunes` algorithm — it highlights the executing source line on the left while the `px`/`sx` pointers and the `*` checkpoint move on the right, with play/pause/scrub and editable pattern/text inputs:

https://gistpreview.github.io/?e52c57571c31e955950e861d25e3bb22

It makes the key property easy to see: only a single checkpoint is kept and grown by one character on each mismatch (O(n·m)), rather than the old recursive fork at every `*`.

pkg/internal/util/wildcard.go
```diff
@@ -15,7 +15,15 @@ limitations under the License.
 */
 
 // !!
-// Modified from https://github.com/minio/minio/blob/RELEASE.2020-06-22T03-12-50Z.hotfix/pkg/wildcard/match.go
+// Originally adapted from https://github.com/minio/minio/blob/RELEASE.2020-06-22T03-12-50Z.hotfix/pkg/wildcard/match.go
```
inteon: This URL does not work for me

wallrj-cyberark: Thanks — you're right, that URL was 404. The original had a bogus `.hotfix` suffix on the tag; the real release tag is `RELEASE.2020-06-22T03-12-50Z`. Fixed in 93811b7:

https://github.com/minio/minio/blob/RELEASE.2020-06-22T03-12-50Z/pkg/wildcard/match.go

While there I also tidied the surrounding note: `minio/minio` is now archived, so I dropped the "still vulnerable" framing and just record the provenance plus why we intentionally diverge (iterative rewrite, `*`-only, rune-based) — and corrected the second link (`minio/pkg` `master` → `main`).

wallrj-cyberark:
This `test` failure looks like an unrelated flake, not a regression from this PR.

- The failing spec is in the readiness controller integration suite (`pkg/internal/controllers/test/ready.go:363`: *"…after enqueue event but for wrong name, should not change returning not-ready"*), which is unrelated to the wildcard matcher this PR changes (`pkg/internal/util/wildcard.go`). The `util` unit tests all passed, and the latest push only edited comments/links.
- The spec `waitForNotReady`s, then flips the fake reconciler to ready (`i++`) and sends an enqueue event **for the wrong name**, asserting via `Consistently` that the policy stays not-ready. The build log shows a *residual* reconcile of the real policy (`reconcileID="7eaa082a-..."`, originating from the initial Create's watch sources) landing just **after** `i++` — it returns `ready=true` and trips the assertion. That reconcile is leftover post-create churn, not the wrong-name event, so the outcome is timing-dependent.
- `test` was green on the previous commit of this branch; only comment/URL text has changed since.

/retest

inteon:
I was not able to find a way to break this algorithm.

/approve
/lgtm

PR #926, opened 2026-06-10
VC-53793: Bound the CEL validator cache with an LRU (alternative to #924)

wallrj-cyberark:
## Summary

Bound the compiled CEL validator cache by replacing the unbounded global
`sync.Map` with a fixed-size LRU (`k8s.io/utils/lru`, already a dependency),
holding at most `DefaultCacheSize` (1024) compiled programs and evicting the
least-recently-used entry when full.

This is an **alternative to #924**, which solves the same problem by tying cache
entries to the CertificateRequestPolicy lifecycle. Both close the
memory-exhaustion variant of the linked issue; this PR does so with a smaller,
self-contained change. See the comparison below.

## Motivation

The CEL validator cache used a global `sync.Map` keyed by expression string with
no eviction — every unique expression compiled during the controller's lifetime
remained in memory permanently. This is the memory-exhaustion variant of
VC-53793 (CWE-770): a principal able to author `CertificateRequestPolicy`
resources — including via `--dry-run=server` CREATEs that are never persisted —
can grow controller memory without bound by submitting unique large expressions.

The CPU/goroutine-hang variant of the same issue was addressed separately by the
CEL cost limit in #923 (already merged, and the base of this branch).

## Design

A bounded cache is preferred here over lifecycle-keyed eviction because it keeps
the cache a self-contained concern:

- No knowledge of policy create/update/delete, and no controller or webhook
  plumbing — `Cache.Get(expr)` is unchanged.
- No per-replica eviction coordination; an entry for a deleted or rewritten
  policy simply ages out.
- A hard memory bound regardless of input.

Trade-off: under a flood of unique expressions, a legitimate policy's compiled
program may be evicted and recompiled on next use. This cost is bounded and is
already capped per compilation by the CEL cost limit from #923.

### Comparison with #924

| Aspect | This PR (LRU) | #924 (per-CRP lifecycle) |
|---|---|---|
| Production code touched | `cache.go` only | `cache.go`, `evaluator.go`, `allowed/validation.go`, `reconciler.go`, controller |
| Cache API | `Get(expr)` unchanged | `Get(policyName, expr)`, `Compile(expr)`, `Remove(policyName)` |
| Memory bound | hard cap (1024 entries) | live policies × expressions |
| New surface | none | `ValidatorCleaner` interface, controller delete-handling |
| Multi-replica eviction | n/a (size-bounded per replica) | best-effort, leader-only on delete |
| New dependency | none (`k8s.io/utils/lru` already present) | none |

## Test plan

- [x] `Test_Cache_Get`: basic caching and error caching
- [x] `Test_Cache_Get_Evicts`: once the size limit is exceeded, the
  least-recently-used entry is evicted and recompiled on next Get
- [x] `Test_Cache_Get_KeepsRecentlyUsed`: touching an entry protects it from
  eviction; the colder entry is evicted instead
- [x] `make verify-golangci-lint` passes (0 issues); `go build ./...` and unit
  tests pass locally; `go mod tidy` produces no go.mod/go.sum changes

### DIFF
--- a/design/20230726-cel-policy.md
+++ b/design/20230726-cel-policy.md
@@ -479,7 +479,31 @@ generous for the simple expressions typical of `CertificateRequestPolicy`
 rules.
 
 The original design also noted that "we should probably consider adding some
-cache expiration logic to avoid appearing like a memory leak." The current
-implementation uses a global `sync.Map` with no eviction or expiry — every
-unique expression compiled during the lifetime of the controller remains in
-memory permanently. Cache lifecycle management is tracked separately.
+cache expiration logic to avoid appearing like a memory leak." The initial
+implementation used a global `sync.Map` keyed by expression string with no
+eviction or expiry — every unique expression compiled during the lifetime of
+the controller remained in memory permanently.
+
+## Addendum: bounded (LRU) validator cache (June 2026)
+
+[#926](https://github.com/cert-manager/approver-policy/pull/926) replaced the
+unbounded global `sync.Map` with a fixed-size least-recently-used cache
+(`k8s.io/utils/lru`, already in the dependency tree), holding at most
+`DefaultCacheSize` (1024) compiled programs. When the limit is exceeded the
+least-recently-used entry is evicted.
+
+This directly closes the memory-exhaustion variant of
+[VC-53793](https://venafi.atlassian.net/browse/VC-53793) (CWE-770): a principal
+able to author policies — including via dry-run CREATEs that are never
+persisted — can no longer grow controller memory without bound by submitting
+unique large expressions, because the cache size is capped regardless of input.
+
+A bounded cache was chosen over tying entries to the CertificateRequestPolicy
+lifecycle because it keeps the cache a self-contained concern. It needs no
+knowledge of policy creation, update or deletion, no controller or webhook
+plumbing, and no per-replica eviction coordination; an entry for a deleted or
+rewritten policy simply ages out. The only trade-off is that, under a flood of
+unique expressions, a legitimate policy's compiled program may be evicted and
+recompiled on next use — a bounded, low cost, and one already capped per
+compilation by the CEL cost limit added in
+[#923](https://github.com/cert-manager/approver-policy/pull/923).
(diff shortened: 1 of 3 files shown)


wallrj-cyberark:
### Note on the `k8s.io/utils/lru` dependency

To save reviewers the investigation: this PR introduces no new dependency — `k8s.io/utils` is already a direct dependency, and `go mod tidy` produces no `go.mod`/`go.sum` change. I read the source so the cache choice can be assessed without digging.

**What it is.** `k8s.io/utils/lru` is a ~99-line wrapper around a vendored copy of Google's groupcache LRU (`internal/third_party/forked/golang/golang-lru`, Apache-2.0) — the canonical `container/list` + `map[interface{}]*list.Element` doubly-linked-list cache.

**Thread-safety (confirmed).** The wrapper holds a `sync.RWMutex`; every mutating method (`Add`, `Get`, `Remove`, `RemoveOldest`, `Clear`) takes the write lock, `Len` takes the read lock. `Get` deliberately takes the **write** lock because a cache hit calls `MoveToFront` and mutates the recency list — a read lock there would be a data race, and they got it right. The forked file is documented "not safe for concurrent access", which is exactly why the locking wrapper exists.

**Correctness (confirmed).** `Add`: existing key → move-to-front + update; new key → push-front then evict oldest if over the limit. `Get`: hit → move-to-front. Both O(1). Standard, no surprises.

**Provenance.** Kubernetes SIG-maintained, wrapping Google's groupcache; both stable for years and widely used across the ecosystem.

**Caveats (none blocking):**
1. The bound is by entry **count** (`DefaultCacheSize` = 1024), not by bytes. The key is the raw expression string and values are compiled programs, so theoretical worst-case memory is ~1024 × (largest accepted expression + program). Still bounded, and the same approach Kubernetes itself uses for compiled-CEL caches; a byte cap would need a different structure and is overkill for this CWE-770 (unbounded *unique* expressions) threat.
2. Every `Get` takes a write lock (inherent to recency-updating LRUs) — fine for this low-contention admission/evaluation path.
3. The `Get`-then-`Add`-on-miss in `cache.Get` is not atomic across the two locked calls, so two goroutines racing on the same new expression may both compile it: harmless duplicate work, no data race, and the same trade-off the previous `sync.Map` implementation made (see the code comment).

PR #945, opened 2026-06-25
fix: enforce policy on full CSR RawSubject

mladen-rusev-cyberark:
RawSubject duplicate-CN / unmapped-OID policy bypass

## TL;DR
A low-privileged tenant (`create certificaterequests` + RBAC `use` on a bound
`CertificateRequestPolicy`) could obtain a CA-signed certificate whose Subject DN
carried identity bits the approver-policy gate never evaluated — a second `CN`, or
unmapped attributes like `emailAddress=` — and impersonate a privileged identity
to first-match Subject parsers (OpenSSL/nginx/HAProxy).

## Root cause
`allowed.Evaluate()` read `csr.Subject` (a `pkix.Name`), which keeps only the
**last** value of a duplicated RDN (e.g. a 2nd CN) and routes unmapped attribute
OIDs (`emailAddress`, `DC`, `UID`, …) into fields the evaluator never reads.
cert-manager copies `csr.RawSubject` into the issued cert, so the gate and the
signer saw different data.

## Fix
Parse `csr.RawSubject` directly and evaluate every value — **fail-closed**:
- every `CommonName` / `SerialNumber` RDN is matched against
  `allowed.commonName` / `allowed.subject.serialNumber`;
- any attribute OID outside the named set is denied unless opted in via the new
  additive `allowed.subject.otherAttributes` field;
- values whose ASN.1 type isn't a supported string encoding
  (`UniversalString`/`GeneralString`, or genuinely non-string `OctetString`/…)
  are denied with actionable guidance to re-issue with a `UTF8String`/
  `PrintableString` subject. Common encodings (incl. `BMPString`/`TeletexString`)
  are matched normally.

Single-CN, named-Subject-field CSRs evaluate exactly as before.

The policy webhook also validates `allowed.subject.otherAttributes` at
**admission** (same as every other `allowed.*` field): a `required` entry must
declare `values`/`validations`, every CEL rule must compile, each `oid` must be a
valid canonical dotted ASN.1 OID, and an `oid` that belongs to a dedicated named
field (e.g. `2.5.4.3`/commonName) is rejected with a pointer to use that field —
preventing dead or impossible-to-satisfy policies.

## Migration impact
| Scenario | After |
|---|---|
| Single CN, named Subject fields | works unchanged |
| Duplicate CN where the 1st is forbidden | denied |
| Subject `emailAddress=` / `DC=` / `UID=` | denied by default — add `allowed.subject.otherAttributes: [{oid: "1.2.840.113549.1.9.1", values: ["*@tenant.example.com"]}]` |
| Non-string / UniversalString Subject value | denied (re-issue as UTF8String) |

Denied requests carry an `Approved=False` condition naming the offending OID/value
and the denying policy; the existing
`certmanager_approverpolicy_certificaterequests_approval{status="denied"}` metric
flags an upgrade-induced rise.


pkg/internal/approver/allowed/evaluator.go
```diff
@@ -104,12 +147,56 @@ type evaluator struct {
...
+}
+
+// subjectValuesForOID returns the string values for every Subject RDN
+// attribute with the given OID across the full RDNSequence (duplicates
+// preserved, unlike the lossy pkix.Name projection). Values whose ASN.1 type
+// does not decode to a Go string (genuinely non-string types, and the rarely
+// used UniversalString/GeneralString encodings) are skipped here — they are
+// denied with an actionable message by the OtherAttributes backstop, so they
+// can never reach this slice unvalidated.
+func subjectValuesForOID(rdns pkix.RDNSequence, oid asn1.ObjectIdentifier) []string {
```
SgtCoDFish: suggestion(AI): I think this finding is fair. The test could be nice, but I think `subjectValuesForOID` could pretty reasonably return `([]string, error)`, with an error if any of the values ends up being a non-string type?

---

Silent coupling: subjectValuesForOID depends on OtherAttributes() to catch its non-string skips

pkg/internal/approver/allowed/evaluator.go:165-178 filters non-string ATVs out of the CN/SerialNumber value list. The safety property that "a non-string CN cannot pass unvalidated" therefore only holds because OtherAttributes() also runs and its walker fires an error on any non-string ATV (evaluator.go:356-362). This coupling is documented in comments but is not defended in code.

If a future maintainer reorders/removes evaluate.OtherAttributes from evaluateFns (evaluator.go:129) — or narrows the walker to unmapped OIDs only, matching what its name suggests — the CN bypass silently reopens.

Suggestions (pick one):
- Add a dedicated pkg/internal/approver/allowed/evaluator_rawbypass_test.go case that stubs OtherAttributes (or removes it from the eval list) and asserts CN with an OctetString is still denied. Any such regression would then fail loudly.
- Or: promote the non-string integrity check out of OtherAttributes into an unconditional first pass that always runs, and drop the redundant if !isString in the OtherAttributes walker.

pkg/internal/approver/allowed/validation.go
```diff
@@ -19,12 +19,26 @@ package allowed
...
 	policyapi "github.com/cert-manager/approver-policy/pkg/apis/policy/v1alpha1"
 	"github.com/cert-manager/approver-policy/pkg/approver"
 )
 
+// validateOID returns a field error (or nil) for a policy-supplied dotted OID
+// string. It must be a valid, canonical ASN.1 OID — evaluation matches it by
+// exact string against the request's parsed OID (always canonical), so a
+// malformed or non-canonical form would silently never match. Canonical means
+// asn1.ObjectIdentifier round-trips to the same string.
+func validateOID(path *field.Path, oid string) *field.Error {
```
SgtCoDFish, line 34: suggestion(AI): This is definitely not required at all, this is gold plating. It's up to you if you want to walk this mile.

---

validateOID accepts negative arcs (defect in upstream helper, exposed here)

pkg/internal/approver/allowed/validation.go:34-40 uses utilpki.ParseObjectIdentifier, which internally calls strconv.Atoi per component. That accepts negatives, so "1.2.3.-4" parses to [1 2 3 -4], String() round-trips to "1.2.3.-4", and the canonical-form guard passes. asn1.ObjectIdentifier values must be non-negative, and X.660 further constrains arc-0 to {0,1,2} with arc-1 range constraints.

Not a security hole (the OID simply won't match a real request), but it lets a dead entry through admission. Cheap fix inline:
for _, c := range parsed {
    if c < 0 {
        return field.Invalid(path, oid, "...")
    }
}
(A stricter arc-0/1 check is nice-to-have; the negative check is the real one.)

---

Further AI context:

In OIDs, an arc is a single node in the OID's hierarchical tree — one dot-separated component of the dotted string.

For 1.2.840.113549.1.9.1 (emailAddress):
- arc-0 = 1
- arc-1 = 2
- arc-2 = 840
- ...and so on

The term comes from the ITU-T X.660 spec, which models the OID namespace as a tree; each edge you traverse from the root is an "arc," and the number labels that arc.

Two structural rules matter for validation:
- arc-0 ∈ {0, 1, 2} — the root has exactly three children: 0 (itu-t), 1 (iso), 2 (joint-iso-itu-t). Anything else is not a real OID.
- arc-1 range depends on arc-0 — under 0 and 1, arc-1 is constrained to 0..39 (a historical encoding quirk: the first two arcs get packed into a single byte as 40*arc0 + arc1). Under 2, arc-1 is unbounded.

pkg/internal/approver/allowed/evaluator.go
```diff
@@ -212,8 +301,113 @@ func (e subjectEvaluator) PostalCode() field.ErrorList {
...
+			// non-string ASN.1 types (OctetString, INTEGER, …) and the
+			// rarely-used UniversalString/GeneralString encodings, which
+			// Go's ASN.1 decoder leaves as a nil interface. The common
+			// string encodings (UTF8String/PrintableString/IA5String/
+			// TeletexString/BMPString) decode to a Go string and are
+			// handled by the named / otherAttributes evaluators. Fail
+			// closed here with actionable guidance so the gate and the
+			// signer agree on what is being issued.
+			if !isString {
+				el = append(el, field.Invalid(
```
SgtCoDFish, line 532: suggestion(AI): Another one that I think is not required. This one is maybe more important because the error would be really confusing... but nobody is going to hit this any time soon, hence unrequired.

---

When a CSR has O=<OctetString>, the denial (evaluator.go:357-361) reports spec.allowed.subject.otherAttributes[2.5.4.10] even though the operator has a spec.allowed.subject.organizations policy. The message text is clear, but the field-path is confusing — the operator is going to look in otherAttributes for 2.5.4.10 and not find it, because it's a named OID.

Cheap improvement: when a non-string ATV has a named OID, prefer the named field's path in the error. E.g., map oidOrganization → subject.organizations. Reuse the same namedSubjectOIDs list.

pkg/internal/approver/allowed/evaluator.go
```diff
@@ -104,12 +147,56 @@ type evaluator struct {
...
+	case 0:
+		// Preserve the existing "required" semantics for the empty case.
+		return e.a.evaluateString(e.request, "", e.allowed.CommonName, fldPath)
+	case 1:
+		return e.a.evaluateString(e.request, values[0], e.allowed.CommonName, fldPath)
+	default:
+		var el field.ErrorList
+		for _, v := range values {
+			el = append(el, e.a.evaluateString(e.request, v, e.allowed.CommonName, fldPath)...)
+		}
```
SgtCoDFish: suggestion(AI): This seems like an easy improvement to the error message?

---

Duplicate CN error paths are indistinguishable

CommonName() case default (evaluator.go:193-199) emits N errors all under spec.allowed.commonName. If both values fail for different reasons (one wildcard-mismatch, one CEL-mismatch), the operator gets two errors at the same path with no ordinal. Consider .Index(i):
el = append(el, e.a.evaluateString(e.request, v, e.allowed.CommonName, fldPath.Index(i))...)

SgtCoDFish (changes requested):
A few AI findings. I think they're valid but I've marked a couple as "definitely not required" because they're quite nitty.

What do you think?

pkg/internal/approver/allowed/evaluator.go
```diff
@@ -212,8 +333,118 @@ func (e subjectEvaluator) PostalCode() field.ErrorList {
...
+// constrain the attribute values as for the named slice fields.
+func (e subjectEvaluator) OtherAttributes() field.ErrorList {
+	fldPath := e.fldPath.Child("otherAttributes")
+
+	// Collect every unmapped-OID attribute, grouped by OID, preserving
+	// the order in which OIDs first appeared in the RDN sequence so
+	// error messages are stable.
+	byOID := make(map[string][]string)
+	var oidOrder []string
+	presentByOID := make(map[string]bool)
```
SgtCoDFish: suggestion: Isn't presentByOID redundant? I think ` _, ok := byOID[key]` is equivalent?

pkg/internal/approver/allowed/evaluator.go
```diff
@@ -104,12 +175,60 @@ type evaluator struct {
...
+func subjectValuesForOID(rdns pkix.RDNSequence, oid asn1.ObjectIdentifier) ([]string, error) {
+	var values []string
+	for _, rdn := range rdns {
+		for _, atv := range rdn {
+			if !atv.Type.Equal(oid) {
+				continue
+			}
+			s, ok := atv.Value.(string)
+			if !ok {
+				return nil, fmt.Errorf("Subject attribute %s uses an unsupported ASN.1 type or string encoding; re-issue the certificate with a UTF8String or PrintableString subject", oid)
```
SgtCoDFish: turbo-nitpick (non-blocking): errors usually start with a lowercase letter in Go, not sure there's a reason to diverge from that here?

pkg/internal/approver/allowed/evaluator.go
```diff
@@ -212,8 +333,118 @@ func (e subjectEvaluator) PostalCode() field.ErrorList {
 	return e.a.evaluateSlice(e.request, e.sub.PostalCode, e.allowed.PostalCodes, e.fldPath.Child("postalCodes"))
 }
 
+// SerialNumber evaluates every Subject SerialNumber RDN value present in
+// csr.RawSubject. As with CommonName the lossy pkix.Name projection keeps
+// only the last value; a duplicate-SN smuggling attempt is policed here.
+// A non-string-encoded SerialNumber is denied directly here (fail-closed).
 func (e subjectEvaluator) SerialNumber() field.ErrorList {
```
SgtCoDFish, line 461: suggestion(AI): I don't think there's a test of multiple serial numbers, is there? Would be worth adding one

pkg/internal/approver/allowed/evaluator.go
SgtCoDFish, line 1: suggestion(AI): This seems like a worthwhile fix to me but definitely not a blocker on its own. AI description below.

---

1. Duplicate error emission for non-string CN / SerialNumber

pkg/internal/approver/allowed/evaluator.go:214-232 (CommonName), :340-358 (SerialNumber), and the OtherAttributes named-OID backstop at :377-395 all fire on the same offending ATV. Trace for a CSR with a non-string CN:

- CommonName() → subjectValuesForOID(..., oidCommonName) returns err → emits one field.Invalid at spec.allowed.commonName with detail "Subject attribute 2.5.4.3 uses an unsupported ASN.1 type or string encoding; ...".
- OtherAttributes() iterates the same RDN, isNamedSubjectOID(CN) == true, non-string branch fires → emits another field.Invalid at spec.allowed.commonName with detail "Subject attribute uses an unsupported ASN.1 type or string encoding; ..." (note: the OID is missing here).

The paths match but the details differ, so field.ErrorList.ToAggregate() won't dedup them. The operator sees two nearly-identical errors for one problem.

Suggestion: in the OtherAttributes backstop, skip CN and SerialNumber explicitly — their dedicated evaluators already cover them. Something like:

```go
if isNamedSubjectOID(atv.Type) {
    // CN and SerialNumber are already covered by their dedicated evaluators
    // (which read from subjectRDNs too). Backstop only fires for the
    // named-slice OIDs (O, OU, L, …) that read the lossy pkix.Name projection.
    if oid := atv.Type; oid.Equal(oidCommonName) || oid.Equal(oidSerialNumber) {
        continue
    }
    if _, ok := atv.Value.(string); !ok {
        ...
    }
    continue
}
```

SgtCoDFish (changes requested):
/lgtm
/approve
/hold

I think all of the comments I've left are ultimately a bit nitpicky. I'd personally vote to make these changes (especially adding the test if it really is missing) but approving with a hold so you can make the call.

pkg/internal/approver/allowed/evaluator.go
```diff
@@ -147,13 +217,61 @@ type evaluator struct {
...
+	var values []string
+	for _, rdn := range rdns {
+		for _, atv := range rdn {
+			if !atv.Type.Equal(oid) {
+				continue
+			}
+			s, ok := atv.Value.(string)
+			if !ok {
+				return nil, fmt.Errorf("subject attribute %s uses an unsupported ASN.1 type or string encoding; re-issue the certificate with a UTF8String or PrintableString subject", oid)
+			}
```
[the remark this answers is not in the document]
mladen-rusev-cyberark: asn1.ObjectIdentifier implements fmt.Stringer (it has a .String() method), so %s renders it as "2.5.4.3" — the dotted OID, exactly as intended. Adding .String() explicitly is harmless but
  unnecessary.

SgtCoDFish (approved):
/lgtm
/approve

I can't see a reason not to merge this. If we need any further changes down the road, happy to review those too!

PR #994, opened 2026-08-21
Adopt the shared kube-api-linter makefile module

wallrj:
Adopts the shared `kube-api-linter` module from cert-manager/makefile-modules#541 (now merged): `klone.yaml` klones the `kube-api-linter` module from makefile-modules `main` (same commit as the other modules), and `.golangci.kube-api-linter.yaml` scopes the linter to `pkg/apis`.

The linter reported 81 findings (the earlier "50" was golangci-lint's default 50-issue cap, which the config now lifts with `max-issues-per-linter: 0`). They are handled in two ways:

- 56 `commentstart` findings are **fixed** (third commit): the godoc now starts with each field's JSON name, mostly via `make fix-kube-api-lint`, plus hand-written godoc for `metadata`, `spec`, `status` and `conditions` which had none. The published CRD field descriptions change accordingly.
- 25 shape/marker findings (`optionalfields`, `arrayofstruct`, `nodurations`, `integers`, `nomaps`, `optionalorrequired`, `conditions`, `ssatags`, `nonpointerstructs`, `markers`) are **suppressed** with `//nolint:kubeapilinter` directives (second commit), following cert-manager/trust-manager#1080: fixing them would change the schema or Go types and could break existing clients; better revisited deliberately for a future API version.

Verified locally: `make verify-kube-api-lint` reports 0 issues and `make generate` / `go build` / `go vet` pass.

/hold

*with claude fable-5*

### DIFF
--- a/.golangci.kube-api-linter.yaml
+++ b/.golangci.kube-api-linter.yaml
@@ -14,3 +14,7 @@ linters:
       - linters:
           - kubeapilinter
         path-except: pkg/apis/.*
+
+issues:
+  max-issues-per-linter: 0
+  max-same-issues: 0
--- a/.golangci.kube-api-linter.yaml
+++ b/.golangci.kube-api-linter.yaml
@@ -8,7 +8,7 @@ linters:
     custom:
       kubeapilinter:
         type: module
-        description: Kube API Linter lints Kube like APIs based on API conventions and best practices.
+        description: Kube API Linter lints Kube-like APIs based on API conventions and best practices.
   exclusions:
     rules:
       - linters:
--- a/.golangci.kube-api-linter.yaml
+++ b/.golangci.kube-api-linter.yaml
@@ -1,0 +1,16 @@
+version: "2"
+
+linters:
+  default: none
+  enable:
+    - kubeapilinter
+  settings:
+    custom:
+      kubeapilinter:
+        type: module
+        description: Kube API Linter lints Kube like APIs based on API conventions and best practices.
+  exclusions:
+    rules:
+      - linters:
+          - kubeapilinter
+        path-except: pkg/apis/.*
--- a/pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
+++ b/pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
@@ -37,12 +37,15 @@ var CertificateRequestPolicyKind = "CertificateRequestPolicy"
 // or denied.
 type CertificateRequestPolicy struct {
 	metav1.TypeMeta `json:",inline"`
+	//nolint:kubeapilinter // commentstart: rewording the godoc would change the published CRD field descriptions
 	// +optional
 	metav1.ObjectMeta `json:"metadata"`
 
+	//nolint:kubeapilinter // commentstart: rewording the godoc would change the published CRD field descriptions
 	Spec CertificateRequestPolicySpec `json:"spec"`
+	//nolint:kubeapilinter // commentstart: rewording the godoc would change the published CRD field descriptions
 	// +optional
-	Status CertificateRequestPolicyStatus `json:"status,omitzero"`
+	Status CertificateRequestPolicyStatus `json:"status,omitzero"` //nolint:kubeapilinter // optionalfields: pre-existing API; changing it now could break clients, revisit for a future API version
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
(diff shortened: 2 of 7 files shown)


pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -37,12 +37,16 @@ var CertificateRequestPolicyKind = "CertificateRequestPolicy"
 // or denied.
 type CertificateRequestPolicy struct {
 	metav1.TypeMeta `json:",inline"`
+	// metadata is the standard object metadata.
 	// +optional
 	metav1.ObjectMeta `json:"metadata"`
 
-	Spec CertificateRequestPolicySpec `json:"spec"`
+	// spec is the desired state of the CertificateRequestPolicy.
+	// +required
```
wallrj, line 45: Self-review: heads-up for reviewers on a near-miss here. `make fix-kube-api-lint` didn't only fix the godoc — its `optionalorrequired` fixer also silently added a `+optional` marker to this previously unmarked field, which dropped `required: [spec]` from the published CRD schema. I missed it at first because I only checked the description changes in the generated CRD diff.

The last commit restores the original schema by marking the field `+required` explicitly (controller-gen had been defaulting it to required because it is a non-pointer field without `omitempty`). The `nonpointerstructs` suppression is needed because KAL would prefer this field to be optional — changing that is a schema change best considered separately.

The generated CRD in this PR is now schema-identical to `main`; only the `description` strings differ.

*with claude fable-5*

.golangci.kube-api-linter.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+linters:
+  default: none
+  enable:
+    - kubeapilinter
+  settings:
+    custom:
+      kubeapilinter:
+        type: module
+        description: Kube API Linter lints Kube like APIs based on API conventions and best practices.
```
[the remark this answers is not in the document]
wallrj: Fixed in 278b75c. This wording is copied from the upstream kube-api-linter README example (https://github.com/kubernetes-sigs/kube-api-linter/blob/092fe0c72997c3e7d6d78405211ed913e37a44e5/README.md?plain=1#L72), so the same "Kube like" text also appears in trust-manager's `.golangci-kal.yml` and the makefile-modules e2e test config — worth fixing upstream at some point.

*with claude fable-5*

pkg/apis/policy/v1alpha1/types_certificaterequestpolicy.go
```diff
@@ -68,25 +72,25 @@ type CertificateRequestPolicySpec struct {
...
 	// CertificateRequest for the request to be allowed by this policy.
 	// Omitted fields place no restrictions on the corresponding
 	// attribute in a request.
 	// +optional
 	Constraints *CertificateRequestPolicyConstraints `json:"constraints,omitempty"`
 
-	// Plugins are approvers that are built into approver-policy at
+	// plugins are approvers that are built into approver-policy at
 	// compile-time. This is an advanced feature typically used to extend
 	// approver-policy core features. This field define plugins and their
```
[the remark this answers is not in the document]
wallrj: Fixed in 278b75c. Both typos are pre-existing (the "This field define plugins" text dates from 93067e7); this PR's autofix only changed the first word of each godoc. Line 101 was a different error — "Omitted fields declares" — also fixed.

*with claude fable-5*

.golangci.kube-api-linter.yaml
```diff
@@ -0,0 +1,20 @@
+version: "2"
+
+linters:
+  default: none
+  enable:
+    - kubeapilinter
+  settings:
```
erikgb, line 7: Did you take a look at the settings we're using in trust-manager? https://github.com/cert-manager/trust-manager/blob/624170596994e291cc433071cd9aaeab88e76564/.golangci-kal.yml#L6-L29

It might be good to align our settings long-term. Maybe we can produce a default KAL config from makefile-modules, as we already do for golangci-lint.

wallrj: Yes — I used the trust-manager config as reference for the issue caps, but deliberately started approver-policy on the KAL defaults to surface the full finding set (81 findings; see the PR description for how they were split between fixes and suppressions).

Agreed on long-term alignment, and the golangci-lint precedent maps well: the `go` module vendors a `.golangci.override.yaml` and yq-merges it over the project config before running ([modules/go/01_mod.mk#L119](https://github.com/cert-manager/makefile-modules/blob/44a7b0216c7b481601f5e80556fc255644bb208d/modules/go/01_mod.mk#L119)). The `kube-api-linter` module could do the same with an agreed default. I've opened cert-manager/makefile-modules#728 to track it — happy to keep this PR on the defaults meanwhile.

*with claude fable-5*

erikgb (commented):
I have reviewed the CRD changes, which I think is the most important in this onboarding PR. And it only seems to change the descriptions of fields. We can try to fix the suppressions in follow-up PRs. Thanks @wallrj! 🚀 

/lgtm
/approve

